Issue #24, opened 2017-01-16
Improve Frontend Development Workflow


fmueller:
At the moment it's quite hard to implement a feature in the swagger-ui frontend or fix a bug their:
- no test data
- OAuth server necessary

Some test data should be provided and it should be easy to disable OAuth for development purpose.

maxim-tschumak:
IMO, it would be easier to add test data using storage API when we have a one-command script (#8) to start the whole application stack. It will also enable us to do proper integration testing locally.

fmueller:
Agreed. I created #31 and close this one.

PR #26, opened 2017-01-17
Enabled running swagger-ui without OAuth integration

maxim-tschumak:
#24 Improving frontend development workflow: 
* Adding readme file including how-to start swagger-ui locally with and without oauth2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ build/
 
 node_modules/
 dist/
+www/
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ dist: trusty
 
 language: node_js
 node_js:
-  - '0.10'
   - '0.12'
 
 addons:
--- a/swagger-ui/Dockerfile
+++ b/swagger-ui/Dockerfile
@@ -4,10 +4,7 @@ FROM registry.opensource.zalan.do/stups/node:0.12-20
 ADD /scm-source.json /scm-source.json
 
 # copy resources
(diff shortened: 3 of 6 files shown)


fmueller:
Can you also add explanation on how to run swagger-ui locally without docker (and without OAuth2)? Ramping up docker container all the time, it quite time consuming in frontend development workflow.

maxim-tschumak:
Good point @fmueller, for development purposes it would be too much overhead to build/start docker every time. Next changes include:
- Build and bundle the resources using Gulp (incl. node server)
- Simplifying Dockerfile (same structure for testing and docker env.)
- Starting the node server as well when `gulp serve`
- Default config doesn't use OAuth
- OAuth integration with `gulp serve-with-oauth`
- Readme about how to start the app locally with/without OAuth integration

maxim-tschumak:
FYI: There is no proper toggle for enabling/disabling OAuth integration in the frontend. Currently, if redirect URL is not set, the user is not being redirected to OAuth server (https://github.com/zalando-incubator/api-discovery/blob/master/swagger-ui/src/main/javascript/SwaggerUi.js#L38).
I think we should implement it in the separate ticket. What do you think, @fmueller?

fmueller:
👍 

Good idea with the oauth toggle. Otherwise it's not possible in frontend to distinguish between wrongly configured OAuth integration and just disabled one.

maxim-tschumak:
ok, thanks for creating an issue :)

Issue #22, opened 2017-01-12
Add zappr integration


fmueller:
Add zappr to this github repository.

PR #28, opened 2017-01-17
Zappr custom rule to accept one approval

maxim-tschumak:
@fmueller thanks for the hint, here is custom zappr config to allow merge after one non-commiter's approval. (#22)


maxim-tschumak:
:+1:

fmueller:
👍

PR #34, opened 2017-01-23
Add OAuth Toggle

### LINKED ISSUE #30: Add toggle for OAuth integration
Currently Swagger UI assumes disabled OAuth integration if redirect url is not set. This can cause issues if OAuth integration is just wrongly configured and should work.

We'd like to have configuration option to enable or disable OAuth integration in Swagger UI explicitly.

See mentions of this here #26

fmueller:
Fixes #30 

Adds:
- Toggle to enabled OAuth explicitly
- Fixes build script

### DIFF
--- a/swagger-ui/README.md
+++ b/swagger-ui/README.md
@@ -23,7 +23,7 @@ gulp serve
 or, with OAuth integration:
 
 ```
-gulp serve serve-with-oauth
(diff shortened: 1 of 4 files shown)


maxim-tschumak:
great! :+1:

Issue #31, opened 2017-01-19
Add test data script


fmueller:
To improve the local development workflow we'd like to have a test script which inserts test data into the storage via its API.

Issue #14, opened 2017-01-10
Remove API exercising


fmueller:
We only wanna use Swagger UI as API Reference Documentation Portal. For that, the API exercising feature causes more trouble than it adds value - the integration with our OAuth2 systems does not work properly all the time. This leads to a poor user experience. Developers should exercise APIs via curl oder more sufficient tooling like Postman.

PR #36, opened 2017-02-07
disabling button and making parameter value input boxes read-only

maxim-tschumak:
Warning: This should be tested in the production environment since we don't have any test data (#31) for local development.
This is a quick-fix for #14.

@fmueller do you think it's enough for now?


maxim-tschumak:
I was able to verify the functionality locally by using test-data from #31 (wip)

fmueller:
👍

Issue #8, opened 2017-01-09
Add one-command build script to start project locally


fmueller:
An one-command build script should be provided which enables users and possible collaborators to start the whole project in Docker - ready to use immediately.

This build script can also reuse docker compose script of #1.

PR #37, opened 2017-02-07
Test data for storage service and run-all script

maxim-tschumak:
This PR covers #31 and #8.

- Test data for storage service
- Script to load test data to storage service
- Run-all script that builds the respective projects, starts docker-compose and loads test data
- Docu adjustments and improvements

@fmueller would be nice if you find some time to test the script on mac.


fmueller:
Tried it out locally.

First I had to install `coreutils` because `realpath` command was not found. `brew install coreutils`fixed this. Then the script runs.

But it failed at the end with: `curl: (7) Failed to connect to 192.168.99.100 port 8010: Connection refused`

192.168.99.100 is my docker-machine ip

maxim-tschumak:
Thank you for checking this!
Our setup defaults to `localhost` as docker host and we forward/open all the ports (swagger-ui, storage and postgres). It looks like the ports might be accessible over localhost (https://docs.docker.com/docker-for-mac/) and the docker host workaround is not needed anymore.

What happens when you just run `./start.sh`?

fmueller:
The script is working. It turns out the sleep interval after docker-compose is not enough on my machine. If I increase the interval it is working.

What about waiting for correct log output like in https://github.com/zalando-incubator/zally/blob/master/.travis.yml#L34 or curl on storage/apps and wait till it returns something?

maxim-tschumak:
Cool, thanks for the hint! Will do so!
Have you tried `./start.sh`?  (If it works with localhost on Mac, we can simplify the start script)

fmueller:
Because you still have to run docker vm with its own ip, it does not work.

maxim-tschumak:
OK, thanks! I'll do the mentioned improvements.

maxim-tschumak:
Update:
- Periodically checking storage service is implemented
- `realpath` is not needed anymore

fmueller:
👍

Issue #4, opened 2017-01-03
Show service url of displayed API definition


fmueller:
It would be great to know the URL of the service exposing the API.

Issue #5, opened 2017-01-03
Show crawled metadata for APIs


fmueller:
It would be great to show some metadata for the APIs. Examples:

- first added
- last API change
- last crawled
- API version

PR #39, opened 2017-02-14
Adding the actual service url

maxim-tschumak:
Implements #4: Show service URL of displayed API definition

@fmueller thanks for the pair programming session!

### DIFF
--- a/swagger-ui/src/main/html/index.html
+++ b/swagger-ui/src/main/html/index.html
@@ -101,7 +101,6 @@
         });
       }
 
-      // just wrap this part also in a function
       function initSwagger(url) {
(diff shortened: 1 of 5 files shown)


maxim-tschumak:
Current Status:
Metadata is not loaded/shown on the initial load. Unfortunately, it's not possible just to fire `update-swagger-ui` event, since this event expects metadata as a parameter.

fmueller:
👍 

@maxim-tschumak I've added metadata loading on initial site load. Can you please validate. Then we can merge this. Further UI beautification can be done with #5

maxim-tschumak:
Functionality has been validated. PR closes #4.
:+1:

PR #43, opened 2017-03-03
Show crawled metadata for apis

maschleg:
implements #5


maxim-tschumak:
:+1:

fmueller:
@maschleg @maxim-tschumak I've seen possible UI issue with existing data we have: service urls sometimes end with trailing slashes. My last commit strips them off to prevent issues in UI when we concatenate it with baseUrl.

fmueller:
👍

Issue #16, opened 2017-01-10
Support multiple environments in which an API is deployed


fmueller:
The storage should support multiple environments where an API is running. Currently an API which runs in environment A and B is stored as API-A and API-B - two separate entities. Instead it should be stored as one API which is deployed in environment A and B. The storage should automatically detect this.

PR #44, opened 2017-03-03
New Meta-API for storage service

maxim-tschumak:
This is the first step for implementing support for multi-environment (#16), it includes new version of the API for the storage service (swagger).
* Support for multiple versions
* Support for multiple applications
* New top-level API resource
* Sub-resources for Version and Application representation


storage/src/main/resources/api/api-discovery-api.yaml
```diff
@@ -0,0 +1,397 @@
...
+    Storage API of the API Discovery Solution (a.k.a Meta-API) is currently used to manage versions and Applications of
+    APIs. Later, it will be also used to manage API's meta information like documentation and user manuals.
+
+    API resource represents an abstract API. One particular API has one or more Versions which can be deployed in
+    multiple Applications. An application can implement multiple APIs.
+
+    APIs, Applications and Versions have a defined lifecycle. Version's lifecycle is based on the usage of the given
+    Version in one or more deployed applications. Application's lifecycle depends on when the application was last seen
+    online. API's lifecycle is an aggregated view on the lifecycles of the linked Versions and Applications (e.g. API
+    is ACTIVE if there is a Version which is deployed in a Application).
```
tfrauenstein: It is unclear (and not trivial to design) what an API entity and life-cycle is and how it relates to (deployed) applications and versions and their life cycles. Related to this is also the question how we can identify these entities. I propose to organise a meeting to clarify the details.

storage/src/main/resources/api/api-discovery-api.yaml
```diff
@@ -0,0 +1,397 @@
...
+      summary: read API
+      description: Returns details about the API
+      operationId: 'org.zalando.apidiscovery.storage.api/read-api'
+      tags:
+      - APIs
+      security:
+      - oauth2:
+        - uid
+      parameters:
+      - name: api_id
```
tfrauenstein: delta API-Id vs API name delivered with endpoint above?

storage/src/main/resources/api/api-discovery-api.yaml
```diff
@@ -0,0 +1,397 @@
...
+  url: https://github.com/zalando-incubator/api-discovery
+
+basePath: /
+produces:
+- application/json
+consumes:
+- application/json
+
+security:
+- oauth2: [uid, overarching.api_definition.write]
```
tfrauenstein: Here we should find better naming than 'overarching' and 'api definition' -- following our new naming concept.
We also should introduce Hostalias definition.

storage/src/main/resources/api/api-discovery-api.yaml
```diff
@@ -0,0 +1,397 @@
...
+      - name: api_id
+        in: path
+        type: string
+        description: ID of the API
+        required: true
+      responses:
+        200:
+          description: API's details
+          schema:
+            $ref: '#/definitions/API'
```
tfrauenstein: why not symmetric to api w/o id, i.e. either 'API' (preferred) or 'APIMetaData' ?

storage/src/main/resources/api/api-discovery-api.yaml
maxim-tschumak: Agree, we should discuss this topic and improve the description. (deterministic finite state machine :) ?)

storage/src/main/resources/api/api-discovery-api.yaml
maxim-tschumak: We discussed the scope name within the team - it should match the new naming concept, but we were also not 100% satisfied with the `overarching.api_definition.write` scope.

storage/src/main/resources/api/api-discovery-api.yaml
maxim-tschumak: So far, the API id is derived from the name in the Swagger definition (human-readable API name in the endpoint above).

storage/src/main/resources/api/api-discovery-api.yaml
maxim-tschumak: It's a trade-off. `GET apis/` endpoint is currently only used to get some basic information about all the APIs (to pre-load the drop-down menu in the frontend). Our intention here is to spare the traffic and provide only needed information - without any embedded sub-resources like versions and applications (Initially, these two endpoints were, in fact, symmetric)

We should discuss this at our to-be-planned meeting too.

maxim-tschumak:
I'm still not 100% sure, if it's a good idea to provide `created` and `last_crawled` in the link relationships between _applications_ and _versions_.
But, as discussed with @fmueller, it's probably the best option for now.

@tfrauenstein, @fmueller please review the changes

storage/src/main/resources/api/api-discovery-api.yaml
maschleg: Good point thats a bit confusing. I would suggest that we return the API resource and filter the subresources by a default include param ``?fields=(name, lifecycle_state)``. See  http://zalando.github.io/restful-api-guidelines/performance/Performance.html

maschleg:
👍

storage/src/main/resources/api/api-discovery-api.yaml
```diff
@@ -0,0 +1,475 @@
...
+            $ref: '#/definitions/Application'
+        versions:
+          type: array
+          items:
+            $ref: '#/definitions/Version'
+
+  APIMetadata:
+    type: object
+    properties:
+      name:
```
tkrop: May be title, if it refers to the swagger title?

PR #47, opened 2017-03-06
Rearrange metadata

### LINKED ISSUE #45: Rearrange metadata information
For readability it is better to have metadata before description text of an API as this can become quite long and as a user of Swagger UI you probably wanna access information like hostname and version faster than reading through or scrolling over description text first.

fmueller:
Fixes #45 

Additionally, I removed some unused code.


maxim-tschumak:
:+1:

PR #49, opened 2017-03-06
Add protocol prefix to service url

### LINKED ISSUE #46: Link service hostname in metadata section correctly
Currently Swagger UI links the specific hostname but doesn't add http or https prefix - so the links are not really clickable in the browser. Depending on swagger spec, Swagger UI should prefix hostname automatically with http or https in metadata section.

fmueller:
Fixes #46


maxim-tschumak:
:+1:

PR #55, opened 2017-03-13
Fix charset handling in crawler

fmueller:
This PR fixes charset handling in crawler: it currently breaks for some services as it does not support UTF-8 in yaml files.


netme:
👍

PR #61, opened 2017-03-15
Fix service url concatenation

### LINKED ISSUE #59: For some APIs Service URL ist not properly concatenated
Storage service creates concatenated service url automatically. This fails for some service urls as they already may have https prefix.

fmueller:
Closes #59


netme:
👍

Issue #66, opened 2017-03-16
Implement Repository API /apis Endpoints 


maschleg:
GET /apis
GET /apis/{api_id}

relates to #16

PR #78, opened 2017-03-17
Implement api resource endpoints

maschleg:
#66 implements GET /apis


storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
```diff
@@ -0,0 +1,49 @@
...
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.toList;
+import static org.zalando.apidiscovery.storage.api.ApiLifecycleState.ACTIVE;
+import static org.zalando.apidiscovery.storage.api.ApiLifecycleState.DECOMMISSIONED;
+import static org.zalando.apidiscovery.storage.api.ApiLifecycleState.INACTIVE;
+
+@Service
+public class ApiService {
+
+    @Autowired
```
fmueller: Favor constructor injection over field based injection. This makes testing without Spring container easier.

fmueller:
👍

PR #79, opened 2017-03-17
Fix api definition of new api

### LINKED ISSUE #72: Fix api definition of new API based on Zally feedback
Zally found some Must violations which we should fix.

fmueller:
Closes #72


maschleg:
👍

PR #91, opened 2017-04-03
Implement api resource endpoints

### LINKED ISSUE #63: Implement Repository API GET /apis/{api_id)
relates to #16

maschleg:
closes #63

### DIFF
--- a/storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDto.java
+++ b/storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDto.java
@@ -8,18 +8,18 @@ import lombok.NoArgsConstructor;
 
 @Data
 @NoArgsConstructor
-public class Api {
+public class ApiDto {
 
     private ApiMetaData apiMetaData;
     private List<VersionsDto> versions;
     private List<ApplicationDto> applications;
 
 
-    public Api(final String name, final ApiLifecycleState lifecycleState) {
+    public ApiDto(final String name, final ApiLifecycleState lifecycleState) {
         apiMetaData = new ApiMetaData(name, lifecycleState);
     }
 
-    public Api(final String name, final ApiLifecycleState lifecycleState, List<VersionsDto> versions, List<ApplicationDto> applications) {
+    public ApiDto(final String name, final ApiLifecycleState lifecycleState, List<VersionsDto> versions, List<ApplicationDto> applications) {
         apiMetaData = new ApiMetaData(name, lifecycleState);
         this.versions = versions;
         this.applications = applications;
--- a/storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDto.java
+++ b/storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDto.java
@@ -2,12 +2,17 @@ package org.zalando.apidiscovery.storage.api;
 
 import java.util.List;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+
+
 @Data
 @NoArgsConstructor
+@JsonInclude(NON_NULL)
 public class ApiDto {
 
     private ApiMetaData apiMetaData;
--- a/storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
+++ b/storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
@@ -17,10 +17,12 @@ import static org.zalando.apidiscovery.storage.api.ApiLifecycleState.INACTIVE;
 public class ApiService {
 
     private final ApiRepository apiRepository;
+    private final ApplicationRepository applicationRepository;
 
     @Autowired
-    public ApiService(ApiRepository apiRepository) {
+    public ApiService(ApiRepository apiRepository, ApplicationRepository applicationRepository) {
         this.apiRepository = apiRepository;
+        this.applicationRepository = applicationRepository;
     }
 
     public List<ApiDto> getAllApis() {
@@ -61,7 +63,7 @@ public class ApiService {
     public Optional<ApiDto> getApi(String apiName) {
         List<ApiEntity> apiEntities = apiRepository.findByApiName(apiName);
 
-        if (!apiEntities.isEmpty()) {
+        if (apiEntities.stream().findFirst().isPresent()) {
             ApiEntity apiEntity = apiEntities.stream().findFirst().get();
 
             return Optional.of(new ApiDto(apiEntity.getApiName(),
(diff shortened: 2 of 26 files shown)


storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDto.java
```diff
@@ -0,0 +1,40 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+
+
+@Data
+@NoArgsConstructor
+@JsonInclude(NON_NULL)
```
fmueller: Would it make sense to configure this project-wide via `JacksonConfiguration`?

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
```diff
@@ -52,10 +54,93 @@ private ApiLifecycleState aggregateApplicationLifecycleStateForDeploymentEntitie
...
 
-    public List<Api> getAllApis(ApiLifecycleState filterByLifecycleState) {
+    public List<ApiDto> getAllApis(ApiLifecycleState filterByLifecycleState) {
         return getAllApis().stream()
             .filter(api -> filterByLifecycleState.equals(api.getApiMetaData().getLifecycleState()))
             .collect(toList());
     }
 
+    public Optional<ApiDto> getApi(String apiName) {
+        List<ApiEntity> apiEntities = apiRepository.findByApiName(apiName);
```
fmueller: I think we should unique constraint to api_name column in database. Then we can rely on this and only return one entity in the query method.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
```diff
@@ -52,10 +54,93 @@ private ApiLifecycleState aggregateApplicationLifecycleStateForDeploymentEntitie
...
+            .findFirst()
+            .map(apiEntity ->
+                Optional.of(new ApiDto(apiEntity.getApiName(),
+                    aggregateApplicationLifecycleStateForApi(apiEntities),
+                    mapVersions(apiEntities),
+                    mapApplications(apiEntities))))
+            .orElse(Optional.empty());
+    }
+
+    private List<VersionsDto> mapVersions(List<ApiEntity> apiEntities) {
```
fmueller: To have so many private mapping methods is an indicator for me that we miss a concept in our codebase so far. Either the mapping should be done in constructor of DTOs itself (if possible) or we introduce some factory classes for this. Both approaches add better testability and readability (hopefully ^^).

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApplicationRepository.java
```diff
@@ -1,6 +1,17 @@
...
+import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.query.Param;
 
 public interface ApplicationRepository extends CrudRepository<ApplicationEntity, String> {
+
+
+    @Query("SELECT DISTINCT a FROM ApplicationEntity a " +
+        "INNER JOIN a.apiDeploymentEntities dep " +
+        "WHERE dep.api in (:apiEntities)")
```
fmueller: Do you know how performant the generated SQL query underneath is? Is it just an IN SQL query or a sub select?

storage/src/test/java/org/zalando/apidiscovery/storage/ApiResourceIntegrationTest.java
```diff
@@ -47,35 +67,21 @@ public void cleanDatabase() {
...
-        ApiEntity testAPi100 = ApiEntity.builder().apiName(TEST_API)
-            .apiVersion("1.0.0")
-            .build();
-        ApiEntity testAPi101 = ApiEntity.builder().apiName(TEST_API)
-            .apiVersion("1.0.1")
-            .build();
-        ApiEntity anotherAPi100 = ApiEntity.builder().apiName(ANOTHER_API)
-            .apiVersion("1.0.0")
-            .build();
+        ApiEntity testAPi100 = givenApiEntity(TEST_API, V1);
```
fmueller: I like this :-) Good idea to add some readable helper methods.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
maxim-tschumak: There could be multiple `api_version` (`ApiEntity`) for one API. I think it's intended to do so by design.

fmueller: Right, mixed API view with implementation view.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
```diff
@@ -52,10 +54,93 @@ private ApiLifecycleState aggregateApplicationLifecycleStateForDeploymentEntitie
...
             .collect(toList());
     }
 
+    public Optional<ApiDto> getApi(String apiName) {
+        List<ApiEntity> apiEntities = apiRepository.findByApiName(apiName);
+
+        return apiEntities.stream()
+            .findFirst()
+            .map(apiEntity ->
+                Optional.of(new ApiDto(apiEntity.getApiName(),
```
maxim-tschumak: We can use `Optional.ofNullable` instead (we don't need the `orElse` clause in this case)

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDto.java
maschleg: Yes, but this would also influences the old API. Therefore we have to adjust some tests. I would like to do it in a seperate task: #93

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
maschleg: Yes thats true I was already thinking about some mapper classes

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApplicationRepository.java
maschleg: It 's an In statement:

``
select distinct applicatio0_.name as name1_3_, applicatio0_.app_url as app_url2_3_, applicatio0_.created as created3_3_ 
from application applicatio0_ inner join api_deployment apideploym1_ on applicatio0_.name=apideploym1_.application_name 
where apideploym1_.api_id in (? , ?)
``
Doesn't look like an expensive query for me. But I could set up the DB locally and run Explain.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDto.java
fmueller: Ok 👍

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApplicationRepository.java
fmueller: Ok, this should be fine then. :)

maschleg:
@fmueller review comments should be fixed now.

maxim-tschumak:
:+1:

PR #92, opened 2017-04-05
New api definition post endpoint

maxim-tschumak:
Implementation of the writing endpoint (`POST /api-definitions`).

Missing parts:
* After #91 is merged, I'll implement the location header (to make use of @maschleg's LinkBuilders)


storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionResourceController.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private CrawlerService crawlerService;
+
+    @Autowired
+    public ApiDefinitionResourceController(CrawlerService crawlerService) {
+        this.crawlerService = crawlerService;
+    }
+
+    @PostMapping
+    public ResponseEntity<?> postCrawledApiDefinition(@RequestBody CrawledApiDefinitionDto crawledAPIDefinitionDto) {
```
fmueller: Minor: What about ResponseEntity<Void>?

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDeploymentEntity.java
```diff
@@ -1,11 +1,6 @@
 package org.zalando.apidiscovery.storage.api;
 
-import javax.persistence.Entity;
-import javax.persistence.EnumType;
-import javax.persistence.Enumerated;
-import javax.persistence.Id;
-import javax.persistence.ManyToOne;
-import javax.persistence.Table;
+import javax.persistence.*;
```
fmueller: Already aligned with @maschleg that we don't wanna have wildcard imports in the codebase.

storage/src/main/java/org/zalando/apidiscovery/storage/api/CrawlerService.java
```diff
@@ -0,0 +1,134 @@
...
+
+import static java.time.OffsetDateTime.now;
+import static java.time.ZoneOffset.UTC;
+
+@Service
+public class CrawlerService {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CrawlerService.class);
+
+    private ApplicationRepository applicationRepository;
```
fmueller: You use `final`all over in the code - but here you missed it. :) If these attributes are not changing, why not adding final to them?

storage/src/main/java/org/zalando/apidiscovery/storage/api/CrawledApiDefinitionDto.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class CrawledApiDefinitionDto {
+
+    /* the current status of the crawling */
```
fmueller: These comments do not add any value.

Maybe for the `version`field where it's not clear which version is meant. But this can and should be solved by better naming, e.g. `apiVersion`.

storage/src/main/java/org/zalando/apidiscovery/storage/api/CrawlerService.java
```diff
@@ -0,0 +1,134 @@
...
+import java.time.OffsetDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import static java.time.OffsetDateTime.now;
+import static java.time.ZoneOffset.UTC;
+
+@Service
+public class CrawlerService {
```
fmueller: Can we find a better name for this class? By looking at the name it's not obvious that it process Api Definitions.

storage/src/main/java/org/zalando/apidiscovery/storage/api/CrawlerService.java
```diff
@@ -0,0 +1,134 @@
...
+        final ApiDeploymentEntity apiDeployment = existingApiDeployment
+                .map(deployment -> deployment)
+                .orElse(newApiDeployment(now));
+
+        apiDeployment.setLastCrawled(now);
+        apiDeployment.setLifecycleState(ApiLifecycleState.ACTIVE);
+        apiDeployment.setApplication(application);
+        apiDeployment.setApi(apiVersion);
+
+        entityManager.persist(apiDeployment);
```
fmueller: I propose to introduce simple Spring data generated repository interfaces for all our JPA entities and would avoid mixing two different abstraction levels for storage: repository and entity manager.

storage/src/main/java/org/zalando/apidiscovery/storage/api/CrawlerService.java
```diff
@@ -0,0 +1,134 @@
...
+
+        final ApplicationEntity application = existingApplication
+                .map(app -> app)
+                .orElse(newApplication(crawledAPIDefinition, now));
+
+        return applicationRepository.saveAndFlush(application);
+    }
+
+    private ApiEntity createOrUpdateApiVersion(final CrawledApiDefinitionDto crawledAPIDefinition, final OffsetDateTime now) {
+        final List<ApiEntity> existingApis = apiRepository.findByApiNameAndApiVersionAndDefinition(
```
fmueller: We wanted to implement this look with a hash of definition itself. What was your reasoning to do it with the full definition content?

storage/src/main/java/org/zalando/apidiscovery/storage/api/CrawlerService.java
```diff
@@ -0,0 +1,134 @@
...
+                          final ApiRepository apiRepository,
+                          final EntityManager entityManager,
+                          final SwaggerDefinitionHelper swaggerHelper) {
+        this.applicationRepository = appRepository;
+        this.apiRepository = apiRepository;
+        this.entityManager = entityManager;
+        this.swagger = swaggerHelper;
+    }
+
+    @Transactional
```
fmueller: We should definitely double-check which transaction isolation level we use to lower the probability of lost updates (which is an open issue in this query-insert-or-update logic).

storage/src/main/java/org/zalando/apidiscovery/storage/utils/SwaggerDefinitionHelper.java
```diff
@@ -0,0 +1,26 @@
...
+import org.codehaus.jackson.JsonNode;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+
+@Component
+public class SwaggerDefinitionHelper {
+
+    private final ObjectMapper objectMapper = new ObjectMapper();
```
fmueller: Am I right that we are not autowiring the `ObjectMapper`here because the ingested swagger definition files have another json field syntax? Because we already configure a global available `ObjectMapper`in `JacksonConfiguration` class.

storage/src/main/resources/api/api-repository-api.yaml
```diff
@@ -75,7 +75,7 @@ parameters:
     description: Filter APIs by lifecycle state
     required: false
   CrawledAPIDefinition:
-    name: CrawledAPIDefinition
+    name: CrawledApiDefinitionDto
```
fmueller: Why did you make this change? I'm confused. ^^ DTO is implementation detail and should not be part of API definition imho. Ya, it's only about naming in this case - but there we should not mix concept and abstractions.

storage/src/main/java/org/zalando/apidiscovery/storage/api/CrawlerService.java
maxim-tschumak: This was my intention too. But it didn't work with the composite id we currently have for this entity. I'll check it again.

storage/src/test/java/org/zalando/apidiscovery/storage/TestDataHelper.java
```diff
@@ -2,10 +2,18 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.stream.Collectors;
+
 import static java.time.OffsetDateTime.now;
 import static java.time.ZoneOffset.UTC;
 
-final class TestDataHelper {
+final public class TestDataHelper {
```
fmueller: Minor: public final class is the default order.

storage/src/main/java/org/zalando/apidiscovery/storage/api/CrawlerService.java
maxim-tschumak: I see multiple ways of implementing this:
- hash index on the definition field
- composite id based on (name,version,definition) (unique index)
- new database field with the hash of the definition
- new database field with the hash of (name,version,definition) which is basically the id of the entity (see the second option)

I prefer to have a unique compound index for (name,version,definition). What do you think, @fmueller @maschleg?

storage/src/main/java/org/zalando/apidiscovery/storage/api/CrawlerService.java
maxim-tschumak: Agree, and I think it's the right scope for the transaction, but will double check this as well.

storage/src/main/java/org/zalando/apidiscovery/storage/utils/SwaggerDefinitionHelper.java
maxim-tschumak: Right, my intention here was to use a clean object mapper. Eventually, when we implement any semantic analysis of the swagger definitions (and have mappings to POJOs in place), we can switch to having a one configured object mapper.

storage/src/main/resources/api/api-repository-api.yaml
maxim-tschumak: My bad, maybe it was an unhappy refactoring. I must definitely check the files I commit :)

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
```diff
@@ -0,0 +1,134 @@
...
+import java.time.OffsetDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import static java.time.OffsetDateTime.now;
+import static java.time.ZoneOffset.UTC;
+
+@Service
+public class ApiDefinitionManager {
```
maschleg: We should align on naming Service/Manager. I prefer Service :)

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private ApplicationEntity createOrUpdateApplication(CrawledApiDefinitionDto crawledAPIDefinition, OffsetDateTime now) {
+        final Optional<ApplicationEntity> existingApplication =
+                applicationRepository.findOneByName(crawledAPIDefinition.getApplicationName());
+
+        final ApplicationEntity application = existingApplication
+                .map(app -> app)
+                .orElse(newApplication(crawledAPIDefinition, now));
+
+        return applicationRepository.saveAndFlush(application);
```
maschleg: Is flushing necessary here?

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
```diff
@@ -0,0 +1,134 @@
...
+        final Optional<ApiDeploymentEntity> existingApiDeployment = Optional.ofNullable(
+                entityManager.find(ApiDeploymentEntity.class, new ApiDeploymentEntity(apiVersion, application)));
+        final ApiDeploymentEntity apiDeployment = existingApiDeployment
+                .map(deployment -> deployment)
+                .orElse(newApiDeployment(now));
+
+        apiDeployment.setLastCrawled(now);
+        apiDeployment.setLifecycleState(ApiLifecycleState.ACTIVE);
+        apiDeployment.setApplication(application);
+        apiDeployment.setApi(apiVersion);
```
maschleg: crawled_state has to be set as well.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
```diff
@@ -0,0 +1,134 @@
...
+
+    void setApiNameAndVersion(final CrawledApiDefinitionDto crawledAPIDefinition) throws SwaggerParseException {
+        final String name;
+        final String version;
+
+        try {
+            name = swagger.nameOf(crawledAPIDefinition.getDefinition());
+            version = swagger.versionOf(crawledAPIDefinition.getDefinition());
+        } catch (IOException | NullPointerException e) {
+            throw new SwaggerParseException("could not parse swagger definition json", e);
```
maschleg: Is this mapped to a 400 status code or how do we handle this?

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
```diff
@@ -0,0 +1,134 @@
...
+    private ApiDeploymentEntity createOrUpdateApiDeployment(ApiEntity apiVersion, ApplicationEntity application,
+                                                            OffsetDateTime now) {
+        final Optional<ApiDeploymentEntity> existingApiDeployment = Optional.ofNullable(
+                entityManager.find(ApiDeploymentEntity.class, new ApiDeploymentEntity(apiVersion, application)));
+        final ApiDeploymentEntity apiDeployment = existingApiDeployment
+                .map(deployment -> deployment)
+                .orElse(newApiDeployment(now));
+
+        apiDeployment.setLastCrawled(now);
+        apiDeployment.setLifecycleState(ApiLifecycleState.ACTIVE);
```
maschleg: this depends on the crawled_state so probably it's not always active.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
maschleg: You also save an existingApplication here.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
```diff
@@ -0,0 +1,134 @@
...
+                .orElse(newApplication(crawledAPIDefinition, now));
+
+        return applicationRepository.saveAndFlush(application);
+    }
+
+    private ApiEntity createOrUpdateApiVersion(final CrawledApiDefinitionDto crawledAPIDefinition, final OffsetDateTime now) {
+        final List<ApiEntity> existingApis = apiRepository.findByApiNameAndApiVersionAndDefinition(
+                crawledAPIDefinition.getApiName(),
+                crawledAPIDefinition.getVersion(),
+                crawledAPIDefinition.getDefinition());
```
maschleg: For performance optimisation I would consider to create a hash for the definition persist it and use it as query param.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
```diff
@@ -0,0 +1,134 @@
...
+
+        LOG.info("New crawling information has been processed; api deployment: {}", apiDeployment);
+    }
+
+    private ApiDeploymentEntity createOrUpdateApiDeployment(ApiEntity apiVersion, ApplicationEntity application,
+                                                            OffsetDateTime now) {
+        final Optional<ApiDeploymentEntity> existingApiDeployment = Optional.ofNullable(
+                entityManager.find(ApiDeploymentEntity.class, new ApiDeploymentEntity(apiVersion, application)));
+        final ApiDeploymentEntity apiDeployment = existingApiDeployment
+                .map(deployment -> deployment)
```
maschleg: this is redundant ;)

maschleg: just ``.orElse(newApiDeployment(now)``

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
```diff
@@ -0,0 +1,134 @@
...
+        entityManager.persist(apiDeployment);
+        return apiDeployment;
+    }
+
+    private ApplicationEntity createOrUpdateApplication(CrawledApiDefinitionDto crawledAPIDefinition, OffsetDateTime now) {
+        final Optional<ApplicationEntity> existingApplication =
+                applicationRepository.findOneByName(crawledAPIDefinition.getApplicationName());
+
+        final ApplicationEntity application = existingApplication
+                .map(app -> app)
```
maschleg: same here

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
maxim-tschumak: I agree + going to rename it.
I would prefer to move all the services to one package and give them meaningful names. (let's discuss this in person)

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
maxim-tschumak: What is `crawled_state` @maschleg?

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
maxim-tschumak: Good point! Thanks for the hint

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
maxim-tschumak: There are different ways to implement it, we should discuss them. (I've mentioned this here: https://github.com/zalando-incubator/api-discovery/pull/92#discussion_r109890436)

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionManager.java
maxim-tschumak: oh, thanks :)

maxim-tschumak:
Thanks for the comments @fmueller @maschleg! I've incorporated the proposed changes. Please have another look, so that we can merge this PR.

storage/src/test/java/org/zalando/apidiscovery/storage/api/APIDefinitionRestIntegrationTest.java
```diff
@@ -0,0 +1,137 @@
...
+import static org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT;
+import static org.springframework.http.MediaType.APPLICATION_JSON_UTF8_VALUE;
+import static org.zalando.apidiscovery.storage.TestDataHelper.invalidCrawledApi;
+import static org.zalando.apidiscovery.storage.TestDataHelper.minimalCrawledApi;
+import static org.zalando.apidiscovery.storage.TestDataHelper.crawlerUberApi;
+
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest(webEnvironment = RANDOM_PORT)
+public class APIDefinitionRestIntegrationTest {
```
maschleg: A Test for the location header would be good :)

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionResourceController.java
```diff
@@ -29,11 +30,15 @@ public ApiDefinitionResourceController(ApiDefinitionProcessingService apiDefinit
...
     public ResponseEntity<Void> postCrawledApiDefinition(@RequestBody CrawledApiDefinitionDto crawledAPIDefinitionDto)
             throws SwaggerParseException {
+        final ApiEntity api = apiDefinitionProcessingService.processCrawledApiDefinition(crawledAPIDefinitionDto);
 
-        apiDefinitionProcessingService.processCrawledApiDefinition(crawledAPIDefinitionDto);
+        final DefinitionDeploymentLinkBuilder linkBuilder = new DefinitionDeploymentLinkBuilder(
+                api.getApiName(),
+                api.getApiVersion(),
+                valueOf(api.getId()));
+        final URI location = URI.create(linkBuilder.buildLink());
```
maschleg: where does the scheme hostname and port come from? An Integration test would be good :)

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -0,0 +1,131 @@
...
+        return apiDeployment;
+    }
+
+    private ApplicationEntity createOrUpdateApplication(CrawledApiDefinitionDto crawledAPIDefinition, OffsetDateTime now) {
+        final Optional<ApplicationEntity> existingApplication =
+                applicationRepository.findOneByName(crawledAPIDefinition.getApplicationName());
+
+        final ApplicationEntity application = existingApplication.orElse(newApplication(crawledAPIDefinition, now));
+
+        return applicationRepository.save(application);
```
maschleg: you still persist existingApplication which have no changes

maxim-tschumak:
Thanks for the hints, @maschleg!

maschleg:
👍

maxim-tschumak:
@fmueller ok for you?

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -0,0 +1,129 @@
...
+                applicationRepository.findOneByName(crawledAPIDefinition.getApplicationName());
+
+        return existingApplication.orElse(applicationRepository.save(newApplication(crawledAPIDefinition, now)));
+    }
+
+    private ApiEntity createOrUpdateApiVersion(final CrawledApiDefinitionDto crawledAPIDefinition, final OffsetDateTime now) {
+        final List<ApiEntity> existingApis = apiRepository.findByApiNameAndApiVersionAndDefinition(
+                crawledAPIDefinition.getApiName(),
+                crawledAPIDefinition.getVersion(),
+                crawledAPIDefinition.getDefinition());
```
maschleg: I aligned with Felix that we should create a hash for the definition and use it as query param for a better performance.

maxim-tschumak:
Still open: refactor the api definitions service in order to avoid transactions.

storage/src/test/java/org/zalando/apidiscovery/storage/api/APIDefinitionRestIntegrationTest.java
```diff
@@ -0,0 +1,150 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT;
+import static org.springframework.http.MediaType.APPLICATION_JSON_UTF8_VALUE;
+import static org.zalando.apidiscovery.storage.TestDataHelper.invalidCrawledApi;
+import static org.zalando.apidiscovery.storage.TestDataHelper.minimalCrawledApi;
+import static org.zalando.apidiscovery.storage.TestDataHelper.crawlerUberApi;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest(webEnvironment = RANDOM_PORT)
+public class APIDefinitionRestIntegrationTest {
```
tkrop: I'm not convinced that mixing up and accessing different layers of a system should be best practice for ITs/UTs?

storage/src/test/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingServiceIntegrationTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT;
+import static org.zalando.apidiscovery.storage.TestDataHelper.crawledMetaApi;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest(webEnvironment = RANDOM_PORT)
+public class ApiDefinitionProcessingServiceIntegrationTest {
```
tkrop: I'm not convinced that mixing up and accessing different layers of a system should be best practice for ITs/UTs?

Besides, this seems to be a good candidate for moving it to a unit test, mocking away components.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -0,0 +1,152 @@
...
+        final ApiDeploymentEntity apiDeployment = createOrUpdateApiDeployment(apiVersion, application, now);
+
+        LOG.info("New crawling information has been processed; api deployment: {}", apiDeployment);
+        return apiVersion;
+    }
+
+    private ApiDeploymentEntity createOrUpdateApiDeployment(ApiEntity apiVersion, ApplicationEntity application,
+                                                            OffsetDateTime now) {
+        final Optional<ApiDeploymentEntity> existingApiDeployment = Optional.ofNullable(
+                entityManager.find(ApiDeploymentEntity.class, new ApiDeploymentEntity(apiVersion, application)));
```
tkrop: This should be handled by the `ApiDeploymentRepotitory`.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -0,0 +1,152 @@
...
+        final ApiDeploymentEntity apiDeployment = existingApiDeployment.orElse(newApiDeployment(now));
+
+        apiDeployment.setLastCrawled(now);
+        apiDeployment.setLifecycleState(ApiLifecycleState.ACTIVE);
+        apiDeployment.setApplication(application);
+        apiDeployment.setApi(apiVersion);
+
+        entityManager.persist(apiDeployment);
```
tkrop: This should be handled by the `ApiDeploymentRepotitory`.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -0,0 +1,152 @@
...
+            throws NoSuchAlgorithmException {
+        final String definitionHash = sha256(crawledAPIDefinition.getDefinition());
+        final List<ApiEntity> existingApis = apiRepository.findByApiNameAndApiVersionAndDefinitionHash(
+                crawledAPIDefinition.getApiName(),
+                crawledAPIDefinition.getVersion(),
+                definitionHash);
+
+        if (existingApis.isEmpty()) {
+            final Session session = entityManager.unwrap(Session.class);
+            final int nextDefinitionId = 1 + (int) session.getNamedQuery("selectLastApiDefinitionId")
```
tkrop: See my comment on the Entity/Query. This should finally be handled by the `ApiDeploymentRepotitory`.

storage/src/main/java/org/zalando/apidiscovery/storage/utils/SwaggerDefinitionHelper.java
```diff
@@ -0,0 +1,26 @@
+package org.zalando.apidiscovery.storage.utils;
+
+import org.codehaus.jackson.JsonNode;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+
+@Component
+public class SwaggerDefinitionHelper {
```
maxim-tschumak: Thanks for pointing this out! The definition is parsed only in the `SwaggerDefinitionHelper`, in controller/service/persistence it is handled as a `String`. Still, it makes sense to improve the parsing within the `SwaggerDefinitionHelper`. I will do so!

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -45,19 +48,36 @@ public ApiDefinitionProcessingService(final ApplicationRepository appRepository,
...
+        session.setFlushMode(FlushModeType.COMMIT);
 
-        apiRepository.save(apiVersion);
+        final ApplicationEntity application = findOrCreateApplication(discoveredApiDefinition, now);
         applicationRepository.save(application);
-        entityManager.persist(apiDeployment);
-        LOG.info("New crawling information has been processed; api deployment: {}", apiDeployment);
-        return apiVersion;
+
+        ApiEntity apiVersion = null;
```
fmueller: @maxim-tschumak this variable can be moved inside the for loop again after our Optional change ;)

maxim-tschumak:
@tkrop please have another look at the PR: requested changes have been incorporated.

maxim-tschumak:
@maschleg @fmueller @tkrop I have a craving for merging this PR :)

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -0,0 +1,177 @@
...
+                        && rootCause.getMessage().toUpperCase().contains("API_VERSION_API_NAME_VERSION_DEFINITION_ID_IDX"))) {
+                    LOG.warn("could not persist discovered api definition {}", e);
+                    throw e;
+                }
+            }
+        }
+        return Optional.empty();
+    }
+
+    private ApiDeploymentEntity findOfCreateApiDeployment(ApiEntity apiVersion, ApplicationEntity application,
```
mfellner: Should this be called findOrCreate...?

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    private ApplicationEntity findOrCreateApplication(DiscoveredApiDefinition discoveredApiDefinition, OffsetDateTime now) {
+        final Optional<ApplicationEntity> existingApplication =
+                applicationRepository.findOneByName(discoveredApiDefinition.getApplicationName());
+
+        return existingApplication.orElse(newApplication(discoveredApiDefinition, now));
+    }
+
+    private ApiEntity findOrCreateApiDefinition(final DiscoveredApiDefinition discoveredApiDefinition, final OffsetDateTime now) {
```
tkrop: I think, this method is implementing two aspects that should be separated:

1. finding out whether an api definition exists for the definition - which should be outside the enclosing store loop.
2. finding out the next definition ID - which is needed inside the existing store loop.

Please separate this methods - a third aspect is calculating the hash that should be outside of both methods.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -0,0 +1,177 @@
...
+
+    protected int nextDefinitionId(DiscoveredApiDefinition discoveredApiDefinition) {
+        return 1 + apiRepository.getLastApiDefinitionId(
+                discoveredApiDefinition.getApiName(),
+                discoveredApiDefinition.getVersion());
+    }
+
+    private String sha256(String content) {
+        messageDigest.update(content.getBytes(StandardCharsets.UTF_8));
+        return String.format("%064x", new BigInteger(1, messageDigest.digest()));
```
mfellner: If we don't call `reset()`, the message digest continues to be initialized with the previous updates. This is probably not the behaviour we want, plus, depending on implementation of the hash, this could leak memory maybe.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionResourceController.java
```diff
@@ -0,0 +1,56 @@
...
+    private ApiDefinitionProcessingService apiDefinitionProcessingService;
+
+    @Autowired
+    public ApiDefinitionResourceController(ApiDefinitionProcessingService apiDefinitionProcessingService) {
+        this.apiDefinitionProcessingService = apiDefinitionProcessingService;
+    }
+
+    @PostMapping
+    public ResponseEntity<Void> postDiscoveredApiDefinition(@RequestBody DiscoveredApiDefinition discoveredAPIDefinition, UriComponentsBuilder builder)
+            throws SwaggerParseException, NoSuchAlgorithmException {
```
tkrop: Please remove `NoSuchAlgorithmException`.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionResourceController.java
```diff
@@ -0,0 +1,56 @@
...
+
+        if (apiOption.isPresent()) {
+            final ApiEntity api = apiOption.get();
+            final LinkBuilder linkBuilder = new DefinitionDeploymentLinkBuilder(
+                    api.getApiName(),
+                    api.getApiVersion(),
+                    valueOf(api.getId()));
+            final URI location = builder.path(linkBuilder.buildLink()).build().encode().toUri();
+            return ResponseEntity.created(location).build();
+        } else {
```
tkrop: Suggestion: Avoid else, if not needed.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionResourceController.java
```diff
@@ -0,0 +1,56 @@
...
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
+        }
+    }
+
+    @ExceptionHandler(SwaggerParseException.class)
+    public ResponseEntity<Void> handleSwaggerParseException(SwaggerParseException e) {
+        return ResponseEntity.badRequest().build();
```
tkrop: What about returning a problem?

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDeploymentEntity.java
```diff
@@ -27,11 +29,11 @@
 public class ApiDeploymentEntity implements Serializable {
 
     @Id
-    @ManyToOne
+    @ManyToOne(cascade = ALL)
```
tkrop: Do we use and need `cascade = ALL` now? Otherwise, may be remove this for now.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDeploymentEntity.java
```diff
@@ -27,11 +29,11 @@
...
     private ApiEntity api;
 
     @Id
-    @ManyToOne
+    @ManyToOne(cascade = ALL)
```
tkrop: Do we use and need `cascade = ALL` now? Otherwise, may be remove this for now.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionResourceController.java
mfellner: The thrown Exceptions can be safely removed here. Do we want to keep them for documentation purposes? At least `NoSuchAlgorithmException` seems very unnecessary.

storage/src/main/resources/api/api-repository-api.yaml
```diff
@@ -75,7 +75,7 @@ parameters:
     description: Filter APIs by lifecycle state
     required: false
   CrawledAPIDefinition:
-    name: CrawledAPIDefinition
+    name: CrawledApiDefinition
```
mfellner: Maybe we should rename this parameter in the whole file?

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -0,0 +1,177 @@
...
+    private final ApiRepository apiRepository;
+    private final EntityManager entityManager;
+
+    @Autowired
+    public ApiDefinitionProcessingService(final ApplicationRepository appRepository,
+                                          final ApiRepository apiRepository,
+                                          final EntityManager entityManager) throws NoSuchAlgorithmException {
+        this.applicationRepository = appRepository;
+        this.apiRepository = apiRepository;
+        this.entityManager = entityManager;
```
tkrop: This is to low level and should not be injected here. Instead we should have an `ApplicationRepository`, and `ApiDefinitionRepository`. Usages should be substituted.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
maxim-tschumak: I've checked that: `MessageDigest.update(byte[])` calls `MessageDigest.engineUpdate(input, 0, input.length)` -> it rewrites the internal state. But, I think it's still better to reset the engine anyways, good point! I'll adjust this.

storage/src/main/resources/api/api-repository-api.yaml
maxim-tschumak: It wasn't intended -> I'll redo this change

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionResourceController.java
maxim-tschumak: Right, thanks for pointing this out!

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
maxim-tschumak: Agree, I'll do this in a separate PR.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiDefinitionProcessingService.java
```diff
@@ -0,0 +1,177 @@
...
+            } catch (DataIntegrityViolationException e) {
+                Throwable rootCause = e.getRootCause();
+                if (!(rootCause instanceof HsqlException
+                        && rootCause.getMessage().toUpperCase().contains("API_VERSION_API_NAME_VERSION_DEFINITION_ID_IDX"))) {
+                    LOG.warn("could not persist discovered api definition {}", e);
+                    throw e;
+                }
+            }
+        }
+        return Optional.empty();
```
maxim-tschumak: agree + will make it consistent to `SwaggerParserException` handling

maxim-tschumak:
@mfellner @tkrop requested changes are implemented, thanks for the comments!
I've also created an issue for repository refactorings: #107.

tkrop:
👍

PR #105, opened 2017-04-12
implement deployments endpoint 

### LINKED ISSUE #104: Implement deployments endpoint
GET /apis/{api_id}/deployments

maschleg:
closes #104


storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiEntity.java
```diff
@@ -1,14 +1,5 @@
 package org.zalando.apidiscovery.storage.api;
 
-import lombok.AllArgsConstructor;
```
fmueller: We should sync our IDE codestyle settings - this is only a rearranging of import statements.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiEntity.java
```diff
@@ -49,4 +49,11 @@
             parameters = {@Parameter(name = "javaZone", value = "UTC")})
     private OffsetDateTime created;
 
+    public void addApiDeploymentEntity(ApiDeploymentEntity apiDeploymentEntity) {
+        if (apiDeploymentEntities == null) {
+            apiDeploymentEntities = new ArrayList<>();
```
fmueller: Could you initialize `apiDeploymentEntities` directly on field declaration with an empty `ArrayList`? Then we can omit this lazy init here.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiResourceController.java
```diff
@@ -1,6 +1,8 @@
 package org.zalando.apidiscovery.storage.api;
 
 
+import java.util.List;
```
fmueller: Again: Import rearrange :)

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
```diff
@@ -118,4 +121,39 @@ public static ApiLifecycleState aggregateApplicationLifecycleStateForDeploymentE
             return Optional.empty();
         }
     }
+
+    public Optional<List<DeploymentDto>> getDeploymentsForApi(String apiId) {
```
fmueller: This pattern is a bit weird: why not returning an empty list instead of an empty Optional?

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
```diff
@@ -118,4 +121,39 @@ public static ApiLifecycleState aggregateApplicationLifecycleStateForDeploymentE
...
+        DefinitionDeploymentLinkBuilder deploymentLinkBuilder = new
+            DefinitionDeploymentLinkBuilder(apiDeploymentEntity.getApi());
+
+        ApplicationDeploymentLinkBuilder applicationDeploymentLinkBuilder = new
+            ApplicationDeploymentLinkBuilder(apiDeploymentEntity.getApplication());
+
+        return DeploymentDto.builder()
+            .apiVersion(apiDeploymentEntity.getApi().getApiVersion())
+            .application(new
+                DeploymentApplicationDto(apiDeploymentEntity.getApplication().getName(),
```
fmueller: Code Style Question: breaking the line after new operator looks strange to me.

storage/src/main/java/org/zalando/apidiscovery/storage/api/DeploymentLinkDto.java
```diff
@@ -47,12 +49,25 @@ public DeploymentLinkDto(ApiDeploymentEntity apiDeploymentEntity, LinkBuilder li
...
+        return getUriComponentsBuilder()
             .pathSegment("applications", applicationName)
             .toUriString();
     }
+
+    private UriComponentsBuilder getUriComponentsBuilder() {
+        if (uriComponentsBuilder != null) {
+            return uriComponentsBuilder;
+        }
+        return UriComponentsBuilder.newInstance();
```
fmueller: Again: What's the benefit of this lazy initialization instead of doing it directly on field declaration?

storage/src/test/java/org/zalando/apidiscovery/storage/ApiDeploymentResourceComponentTest.java
```diff
@@ -0,0 +1,164 @@
...
+    protected ApiRepository apiRepository;
+
+    @Autowired
+    private MockMvc mvc;
+
+    @Configuration
+    @Import(value = {JacksonConfiguration.class, MvcConfiguration.class, ApiResourceController.class})
+    @EnableWebMvc
+    @AutoConfigureDataJpa
+    static class TestConfig {
```
fmueller: For later usage, we should separate this test config and maybe also the test annotation into another class which can be used by other tests easily.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiEntity.java
maschleg: There is a field declaration. But I guess the lombok builder sets the field as null when I don't provide a value (I haven't check the generated code yet, but I got some NPEs).

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiEntity.java
```diff
@@ -49,4 +49,11 @@
             parameters = {@Parameter(name = "javaZone", value = "UTC")})
     private OffsetDateTime created;
 
+    public void addApiDeploymentEntity(ApiDeploymentEntity apiDeploymentEntity) {
+        if (apiDeploymentEntities == null) {
```
tkrop: Why should this be necessary here? You have initialized this `apiDeploymentEntities` in the default constructor (L46). As far as I remember, Hibernate construction can prevent calling it, but I think it is not used here.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiResourceController.java
```diff
@@ -77,6 +77,21 @@ public ApiResourceController(ApiService apiService) {
             .orElse(new ResponseEntity<>(HttpStatus.NOT_FOUND));
     }
 
+    @GetMapping("/{api_id}/deployments")
+    public ResponseEntity<DeploymentsDto> getApiDeployments(@PathVariable("api_id") String apiId,
+                                                            UriComponentsBuilder builder) {
+        return apiService.getDeploymentsForApi(apiId)
+            .map(dtoList -> ResponseEntity.ok(new DeploymentsDto(updateLinks(dtoList, builder))))
+            .orElse(new ResponseEntity<>(HttpStatus.NOT_FOUND));
```
tkrop: Yes, empty list is better here.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApplicationEntityToApplicationDtoConverter.java
```diff
@@ -26,7 +25,7 @@ public static ApplicationDto toApplicationDto(ApplicationEntity applicationEntit
 
     private static DeploymentLinkDto mapApiDeploymentEntityToDefinitionDeploymentLink(ApiDeploymentEntity apiDeploymentEntity) {
         ApiEntity apiEntity = apiDeploymentEntity.getApi();
-        LinkBuilder linkBuilder = new DefinitionDeploymentLinkBuilder(apiEntity.getApiName(), apiEntity.getApiVersion(), valueOf(apiEntity.getId()));
+        LinkBuilder linkBuilder = new DefinitionDeploymentLinkBuilder(apiEntity);
```
tkrop: Codestyle: Shouldn't we inline this variable usage?

storage/src/main/java/org/zalando/apidiscovery/storage/api/DeploymentDto.java
```diff
@@ -0,0 +1,72 @@
...
+
+
+    public void buildLinks(UriComponentsBuilder builder) {
+        definition.buildLink(builder.cloneBuilder());
+        application.buildLink(builder.cloneBuilder());
+    }
+
+    @Data
+    @NoArgsConstructor
+    static class DeploymentApplicationDto {
```
tkrop: I would suggest to skip the redundancy in the name here.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
```diff
@@ -118,4 +121,39 @@ public static ApiLifecycleState aggregateApplicationLifecycleStateForDeploymentE
...
+
+    private List<DeploymentDto> apiEntityToDeploymentDtoList(ApiEntity apiEntity) {
+        return apiEntity.getApiDeploymentEntities().stream()
+            .map(apiDeploymentEntity -> apiDeploymentToDeploymentDto(apiDeploymentEntity))
+            .collect(toList());
+    }
+
+    private DeploymentDto apiDeploymentToDeploymentDto(ApiDeploymentEntity apiDeploymentEntity) {
+        DefinitionDeploymentLinkBuilder deploymentLinkBuilder = new
+            DefinitionDeploymentLinkBuilder(apiDeploymentEntity.getApi());
```
tkrop: This `LinkBuilder`-pattern is absolute over-engineered. I suggest to substitute the whole pattern with a simple domain service using one public function per link type.

storage/src/main/java/org/zalando/apidiscovery/storage/api/DeploymentLinkDto.java
```diff
@@ -47,12 +49,25 @@ public DeploymentLinkDto(ApiDeploymentEntity apiDeploymentEntity, LinkBuilder li
 
     private String applicationName;
 
+    private UriComponentsBuilder uriComponentsBuilder;
+
+    public ApplicationDeploymentLinkBuilder(ApplicationEntity application) {
```
tkrop: See comment above.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiEntity.java
maschleg: It's because of the lombok builder it set's a null value. However I will change this because its too confusing.

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
maschleg: Yes right some functions will propably simplify the link generation.

maschleg:
@fmueller @tkrop please review changes

storage/src/main/java/org/zalando/apidiscovery/storage/api/ApiService.java
```diff
@@ -70,7 +72,7 @@ public static ApiLifecycleState aggregateApplicationLifecycleStateForDeploymentE
         if (!apiEntities.isEmpty()) {
             ApiLifecycleState lifecycleState = aggregateApplicationLifecycleStateForApi(apiEntities);
             List<VersionsDto> versions = toVersionDtoList(apiEntities);
-            List<ApplicationDto> applications = applicationService.getApplicationsByApiEntities(apiEntities);
+            List<org.zalando.apidiscovery.storage.api.ApplicationDto> applications = applicationService.getApplicationsByApiEntities(apiEntities);
```
fmueller: Minor: this should be an import?

tkrop:
👍

PR #108, opened 2017-04-18
New storage endpoint integration

### LINKED ISSUE #69: Crawler should call additionally the new api-definition endpoint /api-definitions/{api_definition_id}

maxim-tschumak:
Closes #69. Implementation is done in the way, that we can easily remove the call to the legacy endpoint afterward.

Includes:
- Implementation of the client for new storage endpoint
- Sequential calls of the legacy and the new endpoint 
- Minor refactorings to make the Job more testable + tests :)


crawler/src/main/java/org/zalando/apidiscovery/crawler/ApiDefinitionCrawlJob.java
```diff
@@ -49,31 +57,50 @@ public Void call() throws Exception {
...
+            legacyStorageClient.createOrUpdateApiDefinition(legacyApiDefinition, app.getId());
+            storageClient.pushApiDefinition(apiDefinition);
         }
+        return null;
+    }
+
+    protected static ApiDefinition constructApiDefinition(JsonNode schemaDiscovery, JsonNode apiDefinition, String schemaType,
+                                                          String appName, String apiDefinitionUrl, String serviceUrl) throws Exception {
+        return ApiDefinition.builder()
+                .status("SUCCESS")
```
fmueller: In the new API it is `SUCCESSFUL`.

maxim-tschumak: Thanks!

crawler/src/main/java/org/zalando/apidiscovery/crawler/storage/LegacyApiDiscoveryStorageClient.java
```diff
@@ -0,0 +1,27 @@
+package org.zalando.apidiscovery.crawler.storage;
+
+import org.springframework.util.Assert;
+import org.springframework.web.client.RestOperations;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class LegacyApiDiscoveryStorageClient {
```
maschleg: ``LegacyApiDiscoveryStorageGateway`` regarding our Rules of Play

crawler/src/main/java/org/zalando/apidiscovery/crawler/storage/LegacyApiDefinition.java
```diff
@@ -0,0 +1,43 @@
...
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class LegacyApiDefinition {
+
+    public static LegacyApiDefinition UNSUCCESSFUL = LegacyApiDefinition.builder().status("UNSUCCESSFUL").build();
+
+    @JsonProperty("status")
```
maschleg: Introduce Jackson Config like in storage pls :)

crawler/src/test/java/org/zalando/apidiscovery/crawler/ApiDefinitionCrawlJobTest.java
```diff
@@ -0,0 +1,130 @@
...
+            .schemaUrl("swagger.json")
+            .serviceUrl("https://meta.api/")
+            .status("SUCCESSFUL")
+            .type("swagger-2.0")
+            .uiLink("/ui")
+            .version("1.0")
+            .build();
+
+    @Test
+    public void shouldBeAbleToConstructApiDefinition() throws Exception {
```
maschleg: Can you please separate the mapping tests and the Job logic.

crawler/src/main/java/org/zalando/apidiscovery/crawler/ApiDefinitionCrawlJob.java
```diff
@@ -19,61 +14,94 @@
...
                 LOG.info("Api definition unavailable for {}", app.getId());
             }
-            return null;
         } catch (Exception e) {
-            storageClient.createOrUpdateApiDefintion(ApiDefinition.UNSUCCESSFUL, app.getId());
             LOG.info("Could not crawl {}: {}", app.getId(), e.getMessage());
-            return null;
+        } finally {
+            legacyStorageClient.createOrUpdateApiDefinition(legacyApiDefinition, app.getId());
+            storageClient.pushApiDefinition(apiDefinition);
```
maschleg: I don't like that this call is depend on a successful remote call of  `` legacyStorageClient.createOrUpdateApiDefinition``. I would wrap at least the first call in an try block. Also we should think about an Retry mechanism (like in hystrix or failsafe) for our gateways.

crawler/src/main/java/org/zalando/apidiscovery/crawler/ApiDefinitionCrawlJob.java
```diff
@@ -19,61 +14,94 @@
...
+import java.io.IOException;
+import java.net.UnknownHostException;
+import java.util.Optional;
+import java.util.concurrent.Callable;
+
 class ApiDefinitionCrawlJob implements Callable<Void> {
 
     private static final Logger LOG = LoggerFactory.getLogger(ApiDefinitionCrawlJob.class);
 
+    private final LegacyApiDiscoveryStorageClient legacyStorageClient;
```
fmueller: Just a thought: Would it be benefitial to seperate legacy and new logic not only on used client class but also on Job class level? So, just spawning two jobs for crawling: the old one (untouched) and new one (new job class).

maxim-tschumak: @fmueller My initial idea was to implement this in different jobs so that don't depend on each other. I decided to go this way to be faster and to be able to remove the call to the legacy endpoint easy. (which we are going to do soon)

fmueller: Ok.

crawler/src/main/java/org/zalando/apidiscovery/crawler/ApiDefinitionCrawlJob.java
```diff
@@ -18,62 +18,103 @@
...
-                        .serviceUrl(serviceUrl)
-                        .schemaUrl(apiDefinitionUrl)
-                        .uiLink(schemaDiscoveryInformation.has("ui_url") ? schemaDiscoveryInformation.get("ui_url").asText() : null)
-                        .definition(apiDefinition.toString())
-                        .build();
-
-                storageClient.createOrUpdateApiDefintion(updateApiDefinitionRequest, app.getId());
+                final JsonNode apiDefinitionInformation = retrieveApiDefinition(serviceUrl + apiDefinitionUrl(schemaDiscoveryInformation));
+
+                legacyApiDefinition = constructLegacyApiDefinition(schemaDiscoveryInformation, apiDefinitionInformation, serviceUrl);
```
tkrop: This and the next conversion should be in `pushApiDefinitionsToLegacyAndNewEndpoint()` or better even up to the gateway.

crawler/src/main/java/org/zalando/apidiscovery/crawler/JacksonConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+
+@Configuration
+public class JacksonConfiguration {
+
+    @Bean
+    public ObjectMapper createObjectMapper() {
+        return new ObjectMapper()
+                .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
+                .configure(SerializationFeature.WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS, false)
+                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```
tkrop: This should allow to get rid of `@JsonIgnoreProperties(ignoreUnknown = true)` - so please remove this annotations.

maxim-tschumak:
@maschleg @fmueller proposed improvements are implemented, do you want to have another look?

crawler/src/main/java/org/zalando/apidiscovery/crawler/JacksonConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+
+@Configuration
+public class JacksonConfiguration {
+
+    @Bean
+    public ObjectMapper createObjectMapper() {
+        return new ObjectMapper()
+                .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
```
tkrop: Please use `disable()` to be more verbose.

crawler/src/main/java/org/zalando/apidiscovery/crawler/ApiDefinitionCrawlJob.java
```diff
@@ -18,62 +18,103 @@
...
             }
-            return null;
         } catch (Exception e) {
-            storageClient.createOrUpdateApiDefintion(ApiDefinition.UNSUCCESSFUL, app.getId());
             LOG.info("Could not crawl {}: {}", app.getId(), e.getMessage());
-            return null;
         }
+
+        pushApiDefinitionsToLegacyAndNewEndpoint(apiDefinition, legacyApiDefinition, app.getId());
+        return null;
```
fmueller: In fact we need a callable here to make use of `invokeAll` and `Futures`. Is there another way to implement this with `Runnable`? Otherwise, we could also return a crawling result instead. ;-)

maxim-tschumak: The used `ExecutorService` expects the jobs to be `Callable`.

maxim-tschumak:
@tkrop please have another look.
changes include:
- clear separation between domain and gateway responsibility
- reducing the usage of exceptions in the business logic

crawler/src/main/java/org/zalando/apidiscovery/crawler/ApiDefinitionCrawlJob.java
```diff
@@ -14,77 +13,60 @@
...
     }
 
-    private Optional<JsonNode> retrieveSchemaDiscovery(String serviceUrl) {
+    private void pushApiDefinitionToLegacyAndNewEndpoint(JsonNode schemaDiscovery, JsonNode apiDefinition, ApplicationBase app) {
+        legacyStorageClient.createOrUpdateApiDefinition(schemaDiscovery, apiDefinition, app);
+        storageClient.pushApiDefinition(schemaDiscovery, apiDefinition, app);
+    }
+
+    private JsonNode retrieveSchemaDiscovery(String serviceUrl) {
```
tkrop: Should be located in the `WellKnownSchemaGateway`.

crawler/src/main/java/org/zalando/apidiscovery/crawler/Utils.java
```diff
@@ -0,0 +1,13 @@
+package org.zalando.apidiscovery.crawler;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public class Utils {
+    public static String extractApiDefinitionUrl(JsonNode schemaDiscovery) {
```
tkrop: You may move this to the `WellKnownSchemaGateway` too.

crawler/src/main/java/org/zalando/apidiscovery/crawler/storage/ApiDiscoveryStorageGateway.java
```diff
@@ -0,0 +1,46 @@
...
+        if (schemaDiscoveryInformation == null || apiDefinitionInformation == null) {
+            apiDefinition = ApiDefinition.UNSUCCESSFUL;
+        } else {
+            apiDefinition = constructApiDefinition(schemaDiscoveryInformation, apiDefinitionInformation, app);
+        }
+
+        restOperations.postForLocation(baseUrl + "/api-definitions", apiDefinition);
+    }
+
+    protected static ApiDefinition constructApiDefinition(JsonNode schemaDiscovery, JsonNode apiDefinition, ApplicationBase app) {
```
tkrop: Add @VisibleForTesting.

crawler/src/test/java/org/zalando/apidiscovery/crawler/TestData.java
```diff
@@ -0,0 +1,60 @@
+package org.zalando.apidiscovery.crawler;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.zalando.apidiscovery.crawler.storage.ApiDefinition;
+import org.zalando.apidiscovery.crawler.storage.LegacyApiDefinition;
+import org.zalando.stups.clients.kio.ApplicationBase;
+
+public class TestData {
```
tkrop: Nice solution. Anyhow, alternate suggestion: put test resources as files in `src/test/resources/`. This has the advantage that resources are slightly better readable and are easier to construct. Drawback only works for resources that are designed for json serialization.

maxim-tschumak:
@tkrop please have another look

crawler/src/main/java/org/zalando/apidiscovery/crawler/gateway/ApiDiscoveryStorageGateway.java
```diff
@@ -0,0 +1,48 @@
...
+        restOperations.postForLocation(baseUrl + "/api-definitions", apiDefinition);
+    }
+
+    @VisibleForTesting
+    protected static ApiDefinition constructApiDefinition(JsonNode schemaDiscovery, JsonNode apiDefinition, ApplicationBase app) {
+        String serviceUrl = app.getServiceUrl().endsWith("/") ? app.getServiceUrl() : app.getServiceUrl() + "/";
+
+        return ApiDefinition.builder()
+            .status("SUCCESSFUL")
+            .type(schemaDiscovery.get("schema_type").asText(""))
```
fmueller: Why do we have empty string fallback for `schema_type` field and not for `title` and `version`?

crawler/src/main/java/org/zalando/apidiscovery/crawler/ApiDefinitionCrawlJob.java
```diff
@@ -1,131 +1,53 @@
...
-            return null;
-        } catch (Exception e) {
-            storageClient.createOrUpdateApiDefintion(ApiDefinition.UNSUCCESSFUL, app.getId());
-            LOG.info("Could not crawl {}: {}", app.getId(), e.getMessage());
-            return null;
+        final JsonNode schemaDiscovery = schemaGateway.retrieveSchemaDiscovery(app);
+
+        if (schemaDiscovery == null) {
+            LOG.info("Api definition unavailable for {}", app.getId());
+            pushApiDefinitionToLegacyAndNewEndpoint(null, null, app);
```
fmueller: `null` values as parameters is not good - hard to understand. You can either provide understandable value objects instead of `null` or provide a new method with an understandable name which is calling this method with corresponding `null` values.

crawler/src/main/java/org/zalando/apidiscovery/crawler/gateway/LegacyApiDefinition.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class LegacyApiDefinition {
+
+    public static LegacyApiDefinition UNSUCCESSFUL = LegacyApiDefinition.builder().status("UNSUCCESSFUL").build();
```
fmueller: I'd like to have `SUCCESS` as constant as well - for symmetry reasons.

crawler/src/main/java/org/zalando/apidiscovery/crawler/gateway/ApiDiscoveryStorageGateway.java
```diff
@@ -0,0 +1,48 @@
...
+
+        restOperations.postForLocation(baseUrl + "/api-definitions", apiDefinition);
+    }
+
+    @VisibleForTesting
+    protected static ApiDefinition constructApiDefinition(JsonNode schemaDiscovery, JsonNode apiDefinition, ApplicationBase app) {
+        String serviceUrl = app.getServiceUrl().endsWith("/") ? app.getServiceUrl() : app.getServiceUrl() + "/";
+
+        return ApiDefinition.builder()
+            .status("SUCCESSFUL")
```
fmueller: We should have this as constant as well.

crawler/src/main/java/org/zalando/apidiscovery/crawler/gateway/WellKnownSchemaGateway.java
```diff
@@ -0,0 +1,103 @@
...
+import org.springframework.web.client.ResourceAccessException;
+import org.springframework.web.client.RestTemplate;
+import org.zalando.stups.clients.kio.ApplicationBase;
+
+import java.io.IOException;
+import java.net.UnknownHostException;
+
+public class WellKnownSchemaGateway {
+
+    private static final Logger LOG = LoggerFactory.getLogger(WellKnownSchemaGateway.class);
```
fmueller: Since we are using Lombok, we should use `@Slf4j` annotation on class level to have `log`attribute available. No need to declare it anymore.

crawler/src/main/java/org/zalando/apidiscovery/crawler/gateway/WellKnownSchemaGateway.java
```diff
@@ -0,0 +1,103 @@
...
+    protected static String extractApiDefinitionUrl(JsonNode schemaDiscovery) {
+        String apiDefinitionUrl = schemaDiscovery.get("schema_url").asText();
+        if (apiDefinitionUrl.startsWith("/")) {
+            apiDefinitionUrl = apiDefinitionUrl.substring(1);
+        }
+        return apiDefinitionUrl;
+    }
+
+    @VisibleForTesting
+    protected static String extractServiceUrl(ApplicationBase app) {
```
fmueller: Does this method have to be static for testing?

crawler/src/main/java/org/zalando/apidiscovery/crawler/gateway/WellKnownSchemaGateway.java
```diff
@@ -0,0 +1,103 @@
...
+            } else {
+                LOG.info("Try to load api definition as json for service {}", app.getId());
+                return tryRetrieveApiDefinitionAsYaml(url, app.getId());
+            }
+        } catch (Exception e) {
+            return tryRetrieveApiDefinitionAsYaml(url, app.getId());
+        }
+    }
+
+    protected static String extractApiDefinitionUrl(JsonNode schemaDiscovery) {
```
fmueller: Does this have to be static and is a `@VisibleForTesting` annotation missing here?

crawler/src/main/java/org/zalando/apidiscovery/crawler/gateway/WellKnownSchemaGateway.java
```diff
@@ -0,0 +1,103 @@
...
+
+        HttpHeaders headers = new HttpHeaders();
+        headers.add("Accept", "*/*");
+        ResponseEntity<String> yamlApiDefinition = restTemplate.exchange(url, HttpMethod.GET, new HttpEntity(headers), String.class);
+
+        if (!yamlApiDefinition.getStatusCode().is2xxSuccessful()) {
+            LOG.info("Could not load yaml api definition");
+            return null;
+        }
+        return new ObjectMapper(new YAMLFactory()).readValue(yamlApiDefinition.getBody(), JsonNode.class);
```
fmueller: Would it make sense to instantiate this `ObjectMapper` for YAML mapping as class level attribute only once?

crawler/src/main/java/org/zalando/apidiscovery/crawler/gateway/WellKnownSchemaGateway.java
maxim-tschumak: It doesn't have to be static, but what is the benefit of making this method non-static? This method can be used without having an instance of the `WellKnownSchemaGateway`.

crawler/src/main/java/org/zalando/apidiscovery/crawler/gateway/WellKnownSchemaGateway.java
maxim-tschumak: Yes :+1:

crawler/src/main/java/org/zalando/apidiscovery/crawler/ApiDiscoveryCrawler.java
```diff
@@ -8,62 +8,64 @@
...
                 .collect(Collectors.toList());
-        LOG.info("Crawling {} api definitions", crawlJobs.size());
+        log.info("Crawling {} api definitions", crawlJobs.size());
 
         try {
-            List<Future<Void>> futures = fixedPool.invokeAll(crawlJobs);
-            for (Future<Void> future : futures) {
+            List<Future<CrawlResult>> futures = fixedPool.invokeAll(crawlJobs);
+            for (Future<CrawlResult> future : futures) {
                 future.get();
```
fmueller: And later we use information from `'CrawlResult` for extending error state in storage, right? :)

maxim-tschumak: right :)

maxim-tschumak:
@fmueller @tkrop please have another look

crawler/src/main/java/org/zalando/apidiscovery/crawler/KioApplication.java
```diff
@@ -0,0 +1,28 @@
...
+    public String getServiceUrl() {
+        return app.getServiceUrl().endsWith("/") ? app.getServiceUrl() : app.getServiceUrl() + "/";
+    }
+
+    public String getName() {
+        return app.getId();
+    }
+
+    @Override
+    public boolean equals(Object object) {
```
fmueller: Why no `@EqualsAndHashCode` annotation in this case? And if you override `equals` you also have to override `hashCode` to fulfill the contract.

maxim-tschumak: `@EqualsAndHashCode` doesn't work since `ApplicationBase` doesn't have a proper `equals`/`hashCode` implementation. I'll add `hashCode` implementation to `KioApplication`.

fmueller:
Ok, can we file an issue for the kio client library? Because I don't like using `toString` in `equals` and `hashCode`. ;-)

👍

maxim-tschumak:
@fmueller I'll do it tomorrow :)

maxim-tschumak:
@fmueller I've created an issue for `kio-client-java`: https://github.com/zalando-incubator/kio-client-java/issues/13

PR #112, opened 2017-04-24
Update versions of crawler dependencies

fmueller:
This PR updates versions of dependencies of the crawler, only minor library updates.


maxim-tschumak:
:+1:

PR #117, opened 2017-04-27
#111 one instead of two calls to get api definition

### LINKED ISSUE #111: Don't call api definition endpoint twice
**Current behavior**
Crawler calls api definition endpoint twice: first trying to parse `JsonNode`, if this fails, it tries to parse as `String` and convert this from YAML to JSON afterwards. But for each attempt a network call is issued - and each network call accepts all content types.

**Expected behavior**
Only one network call is issued per api definition endpoint of a given service. Afterwards the crawler tries to parse this as JSON or YAML.

**Depends on** #69

maxim-tschumak:
Fixes #111.

PR includes:
- only one call to get api definition
- expecting `String` and then parsing it to JSON/YAML
- no retry on `RestClientException`


crawler/src/test/java/org/zalando/apidiscovery/crawler/gateway/WellKnownSchemaGatewayTest.java
```diff
@@ -98,29 +94,22 @@ public void shouldSuccessfullyRetrieveApiDefinition() throws Exception {
...
-        doThrow(new RestClientException("oh no!")).when(restTemplate).exchange(
-            eq("https://meta.api/swagger.json"),
-            eq(HttpMethod.GET),
-            any(HttpEntity.class),
-            eq(JsonNode.class));
-
-        doReturn(new ResponseEntity<>("some: yaml", HttpStatus.OK)).when(restTemplate).exchange(
+    public void shouldRetrieveYamlApiDefinition() throws Exception {
+        JsonNode expectedApiDefinition = readJson(metaApiDefinitionResource);
+        SchemaDiscovery schemaDiscovery = new SchemaDiscovery(readJson(metaApiSchemaDiscoveryResource));
```
fmueller: We should improve this and make clear that you are reading and validating YAML here:

- `readJson` is misleading
- please place a YAML file as `Resource` like you did with the JSON examples
- if you request the file, can you also do it with `.yaml` ending (I know, this does not make a huge difference, but it mimics the expected production behavior best

crawler/src/main/java/org/zalando/apidiscovery/crawler/gateway/WellKnownSchemaGateway.java
```diff
@@ -55,36 +54,19 @@ public JsonNode retrieveSchemaDiscovery(KioApplication app) {
...
-    }
-
-    @VisibleForTesting
-    protected JsonNode tryRetrieveApiDefinitionAsYaml(String url, String appId) throws IOException {
-        log.info("Try to load api definition as yaml for service {}", appId);
 
-        HttpHeaders headers = new HttpHeaders();
-        headers.add("Accept", "*/*");
-        ResponseEntity<String> yamlApiDefinition = restTemplate.exchange(url, HttpMethod.GET, new HttpEntity(headers), String.class);
+        final HttpHeaders headers = new HttpHeaders();
```
fmueller: Minor: What about extracting this into `acceptAllHeader()` returning properly configured `HttpHeaders` object. Then we could inline this call in `restTemplate` call.

fmueller:
👍

PR #121, opened 2017-05-05
Improving package structure

### LINKED ISSUE #97: Reasonable package structure
The storage project is getting bigger. We should move some classes around and create a good package structure.

maxim-tschumak:
Closes #97.

Please have a look at the new package structure. It has become a mixture of the one defined in our [Rules of Play](https://pages.github.bus.zalan.do/meta-api/rules-of-play/architecture/) and [Microservice Anatomy](https://martinfowler.com/articles/microservice-testing/#anatomy-modules) by Martin Fowler and Co.


maschleg:
Package structure overview:
```
storage/src/main/java
└── org
    └── zalando
        └── apidiscovery
            └── storage
                ├── api
                │   ├── domain
                │   │   ├── logic
                │   │   └── model
                │   ├── repository
                │   └── resource
                └── legacy
```

mfellner:
I just realised that "domain" can be its own package (group) or the packages (groups) "repository", "domain" and "service". How are we going to treat "domain" exactly?

maxim-tschumak:
@mfellner 

> Maybe we don't need the toplevel api package? It seems it's purpose is only to separate the "new" code from the "legacy" code.

Yes :) , it's temporary, we will move the new production code to `storage/src/main/java/org/zalando/apidiscovery/storage`

> In any case "service" might be a good package to have for all the service classes. Maybe "logic" should actually be "service"?

It was my initial idea too. I'm OK with both solutions.

> The domain package includes logic and model, however not repository. According to Fowler all three are "domain" however. We could put all three packages into "domain" or get rid of "domain" and put the packages in the root and just consider them "domain".

I think my second/third commit implements exactly the structure you have described. But, I would prefer to have an explicit domain package, since it's the core and also the biggest part of the service. But again, I'm OK with both approaches.

> I just realised that "domain" can be its own package (group) or the packages (groups) "repository", "domain" and "service". How are we going to treat "domain" exactly?

I propose to use Martin Fowler's definition. Package structure should differ on one point though: `repository` is on the same structural level as `resource` and `domain`. The reason for this is the fact that our `repository` units are a mixture of DataMappers, ORM and Domain/Repositories in Martin Fowler's terminology. And, intuitively, they belong together.

@maschleg @fmueller @tkrop what do you think?

maschleg:
1. I am ok with a temporary api / legacy package 
2. I would call it also service, because it's more common.
3. I prefer explicit domain package 
4. I would rename the package repository to persistence.

fmueller:
Just to add my 2 cents: I like package structure by vertical slices instead of layers, see http://www.markhneedham.com/blog/2012/02/20/coding-packaging-by-vertical-slice/

maxim-tschumak:
@fmueller @tkrop @mfellner @maschleg adjusted to the discussed package structure

fmueller:
👍

Issue #118, opened 2017-04-28
Upgrade to swagger-ui 3.x


mfellner:
### What

Upgrade to [swagger-ui 3.x](https://github.com/swagger-api/swagger-ui) and rewrite server and build chain.

### Why

The current implementation of the api-discovery frontend is based on an outdated version of swagger-ui and a somewhat hacked-together server combined with a complicated set of build scripts.

This makes the project hard to maintain and improve on, plus we're missing out on new features in upcoming versions of swagger-ui.

### Proposal

* Create a webpack based frontend project
* Pull in the latest swagger-ui as an npm package
* Rewrite the server

mfellner:
Closed with #127

PR #127, opened 2017-05-18
Add Swagger-UI 3

mfellner:
Part of #118 

* Add new client based on Swagger UI 3.0
* Add server for serving static files and for API delegation + authentication


swagger-ui-3/README.md
```diff
@@ -0,0 +1,29 @@
+# API Discovery UI
+
+Based on [swagger-ui 3](https://github.com/swagger-api/swagger-ui).
+
+### Configuration
+
+In your browser, set an OAUTH token for the local storage value `API_DISCOVERY_TOKEN`.
+
+Then you can load new API definitions by typing in a valid Api Discovery Storage service app-endpoint URL and pressing 'Discover'. E.g. `https://apidisco.zalando.com/apps/awesome-app`.
```
maschleg: I am not sure If I understand this correctly. Wouldn't it make more sense to provide the storage url as config variable cause for our use case it´s always the same?

maxim-tschumak: @maschleg I believe it's just for testing purposes (this PR is WiP)

We should discuss how users' navigation should be implemented here.

mfellner: @maschleg @maxim-tschumak The README probably sounds confusing because this solution is still a WIP. Setting an OAUTH token in the browser and manually entering the URL are just stopgaps for development.

The final solution should look like this, in my opinion:

* all configuration is managed by the server (NodeJS backend) and client configuration is sent to the client by the server (by sending it on page load, e.g. in HTML)
* communication to other backend services (e.g. API storage) is managed by the server as well (via service2service authentication), the client talks to a facade/proxy API of the server
* user authentication happens between the client and the server via OAUTH2 implicit flow
* all APIs in the storage service are loaded and cached by the server and are made available in the client and discoverable through the UI

maxim-tschumak: Additionally, we have to think about what best user experience can look like: e.g. how does the user find an API definition. (search by API name and/or application name, full-text search, a drop-down menu for the semantic versions of a particular API, etc.)

maxim-tschumak:
:+1:

mfellner:
👍

PR #128, opened 2017-05-19
K8s iam integration

### LINKED ISSUE #126: Upgrade tokens and spring-oauth2-client library to support k8s IAM integration
## Task

update: 
``org.zalando.stups:stups-spring-oauth2-client``
``org.zalando.stups:tokens``
to support k8s IAM integration.

Move tokens configuration from application to environment context.

maschleg:
closes #126


crawler/src/main/java/org/zalando/apidiscovery/crawler/OAuthConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+
+@Configuration
+public class OAuthConfiguration {
+
+    @Value("${tokens.access-token-uri}")
+    private String accessTokenUri;
```
fmueller: If this is only used in `accessTokens` we can make it a method parameter.

crawler/build.gradle
```diff
@@ -38,8 +38,8 @@ dependencies {
     compile('org.springframework.boot:spring-boot-starter')
     compile("org.springframework.boot:spring-boot-starter-actuator")
     compile('org.apache.httpcomponents:httpclient:4.5.3')
-    compile('org.zalando.stups:stups-spring-oauth2-client:1.0.18')
-    compile('org.zalando.stups:spring-boot-zalando-stups-tokens:0.9.11')
+    compile('org.zalando.stups:stups-spring-oauth2-client:1.0.19')
+    compile('org.zalando.stups:tokens:0.11.0-beta-2')
```
fmueller: Why is this only beta? Feels weird at first view.

crawler/src/main/java/org/zalando/apidiscovery/crawler/OAuthConfiguration.java
maschleg: We don't need this add all here.

crawler/build.gradle
maschleg: I know I had the same doubt, but I talked to Joerg Bellman and he says it's stable for production.

maschleg:
@fmueller please review changes/comments

fmueller:
👍

PR #131, opened 2017-05-24
Validate required fields post api definitions

### LINKED ISSUE #130: Validate required fields for POST /api-definitions
## Task

Validate required fields for POST /api-defintions.

If a required field is not provided return a 404 Bad Request with an error description.  

depends on #129

maschleg:
closes #130


maxim-tschumak:
:+1:

PR #144, opened 2017-06-06
Add lifecycle management to API Repository

### LINKED ISSUE #123: Adapt Lifecycle Management to new API Repository API
The new API Repository does not implement automatic lifecycle management. The `ApiLifecycleManager` class has to be adapted for the new  data structures.

fmueller:
Closes #123

### DIFF
--- a/storage/src/main/java/org/zalando/apidiscovery/storage/domain/service/ApiLifecycleService.java
+++ b/storage/src/main/java/org/zalando/apidiscovery/storage/domain/service/ApiLifecycleService.java
@@ -1,0 +1,77 @@
+package org.zalando.apidiscovery.storage.domain.service;
+
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Transactional;
+import org.zalando.apidiscovery.storage.domain.model.ApiLifecycleState;
+import org.zalando.apidiscovery.storage.repository.ApiEntity;
+import org.zalando.apidiscovery.storage.repository.ApiRepository;
+
+import java.time.OffsetDateTime;
+import java.util.List;
+
+import static java.time.OffsetDateTime.now;
+import static java.time.ZoneOffset.UTC;
+
+/**
+ * This component schedules a recurring job which checks for lifecycle state changes:
+ * <ul>
+ * <li>if an API deployment was not crawled successfully for certain amount of time mark it as inactive</li>
+ * <li>if an API deployment was not crawled successfully even longer mark it as decommissioned</li>
+ * </ul>
+ */
+@Component
+@Slf4j
+class ApiLifecycleService {
+
+    private final ApiRepository apiRepository;
+    private final int markAsInactiveTime;
+    private final int markAsDecommissionedTime;
+
+    @Autowired
+    ApiLifecycleService(ApiRepository apiRepository,
+                        @Value("${inactive.time}") int markAsInactiveTime,
+                        @Value("${decommissioned.time}") int markAsDecommissionedTime) {
+        this.apiRepository = apiRepository;
+        this.markAsInactiveTime = markAsInactiveTime;
+        this.markAsDecommissionedTime = markAsDecommissionedTime;
+    }
+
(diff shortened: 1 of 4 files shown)


fmueller:
The scope violations in Codacy as false positives in my opinion as package scope is the least possible visibility for this class as it needs Spring and its component scan features.

storage/src/main/java/org/zalando/apidiscovery/storage/domain/service/ApiLifecycleService.java
```diff
@@ -0,0 +1,77 @@
...
+    @Scheduled(fixedDelayString = "${lifecycle-check.delay}")
+    void checkLifecycleStates() {
+        final OffsetDateTime now = now(UTC);
+        inactivateApis(now);
+        decommissionApis(now);
+    }
+
+    @Transactional
+    void inactivateApis(OffsetDateTime now) {
+        final OffsetDateTime toOldApis = now.minusSeconds(markAsInactiveTime);
```
maxim-tschumak: minor: probably `tooOldApis`

storage/src/main/java/org/zalando/apidiscovery/storage/domain/service/ApiLifecycleService.java
```diff
@@ -0,0 +1,77 @@
...
+
+        List<ApiEntity> inactivatedApis = apiRepository.findNotUpdatedSinceAndActive(toOldApis);
+        inactivatedApis.forEach(a -> a.getApiDeploymentEntities().stream()
+            .filter(d -> d.getLifecycleState() == ApiLifecycleState.ACTIVE
```
maxim-tschumak: The filters you apply here on the results of `findNotUpdatedSinceAndActive` are redundant: these criteria are already specified in the SQL query.

storage/src/main/java/org/zalando/apidiscovery/storage/domain/service/ApiLifecycleService.java
```diff
@@ -0,0 +1,77 @@
...
+        apiRepository.save(inactivatedApis);
+    }
+
+    @Transactional
+    void decommissionApis(OffsetDateTime now) {
+        final OffsetDateTime toOldApis = now.minusSeconds(markAsDecommissionedTime);
+
+        List<ApiEntity> decommissionedApis = apiRepository.findNotUpdatedSinceAndInactive(toOldApis);
+        decommissionedApis.forEach(a -> a.getApiDeploymentEntities().stream()
+            .filter(d -> d.getLifecycleState() == ApiLifecycleState.INACTIVE
```
maxim-tschumak: Same here, we don't have to double check the attributes.

storage/src/main/java/org/zalando/apidiscovery/storage/domain/service/ApiLifecycleService.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Scheduled(fixedDelayString = "${lifecycle-check.delay}")
+    void checkLifecycleStates() {
+        final OffsetDateTime now = now(UTC);
+        inactivateApis(now);
+        decommissionApis(now);
+    }
+
+    @Transactional
```
maxim-tschumak: Does the method have to be transactional?

storage/src/main/java/org/zalando/apidiscovery/storage/domain/service/ApiLifecycleService.java
fmueller: I doubled checked it: we have to, as JPA loads the whole list of API deployments for an API regardless of the query to fetch a list of APIs.

storage/src/main/java/org/zalando/apidiscovery/storage/domain/service/ApiLifecycleService.java
fmueller: Fixed this, thanks.

storage/src/main/java/org/zalando/apidiscovery/storage/domain/service/ApiLifecycleService.java
fmueller: I would prefer to have it here since we operate on two tables: `api_version` and `api_deployment`. In the meantime new API Definition request can change lifecycle status, for instance.

fmueller:
Please re-check @maxim-tschumak :)

maxim-tschumak:
:+1:

maxim-tschumak:
:+1:

PR #145, opened 2017-06-06
Fix missing OAuth scope check for posting new api definitions

### LINKED ISSUE #136: POST /api-definitions not secured with scope
**Current behavior**
new Api definitions can be created with any valid token.

**Expected behavoir**
POST /api-definitions  is secured with a scope (application.write_all).

fmueller:
Closes #136


maxim-tschumak:
:+1:

PR #146, opened 2017-06-06
Add validation for blank or missing api name field

### LINKED ISSUE #140: Database field api_version.api_name must not blank
**Current behavior** 
It is possible to persist an api_version with an blank api_name. As api_name is part of the unique constraint this must be prevent.  

**How to repoduce**
Post /api-defintions with an empty title in the definition. 

**Expected behavior** 

Solution must be clarified: 
1. Post /api-defintions returns a Bad Request
2. We have mechanism to persist invalid api definitions

fmueller:
Closes #140


maxim-tschumak:
:+1:

PR #147, opened 2017-06-07
#126 removing actuator

maxim-tschumak:
`org.springframework.boot:spring-boot-starter-actuator` requires `org.springframework.boot:spring-boot-starter-web`. We decided not to use Actuator and have another health check in place.


maschleg:
👍

PR #148, opened 2017-06-07
#6 adding monitoring and health check endpoint

maxim-tschumak:


crawler/build.gradle
```diff
@@ -36,6 +36,7 @@ repositories {
 
 dependencies {
     compile('org.springframework.boot:spring-boot-starter')
+    compile("org.springframework.boot:spring-boot-starter-web")
```
fmueller: Minor: I think with starter-web in place you can remove starter.

fmueller:
👍

crawler/build.gradle
maxim-tschumak: Thanks, I've removed the unused dependency.

fmueller:
👍

PR #150, opened 2017-06-08
Fix /apis endpoint

### LINKED ISSUE #149: Incorrect /apis endpoint
The `/apis` endpoint implemented by [ApiResourceController.java#L42](https://github.com/zalando-incubator/api-discovery/blob/master/storage/src/main/java/org/zalando/apidiscovery/storage/resource/ApiResourceController.java#L42) does not match the API definition at [api-repository-api.yaml#L107](https://github.com/zalando-incubator/api-discovery/blob/master/storage/src/main/resources/api/api-repository-api.yaml#L107).

* Instead of a list of `ApiMetaData`, the returned ApiList object contains a list of `Api`.
* `ApiMetaData` lacks the `id` property.

mfellner:
Closes #149.

* ApiList.apis is now a collection of ApiDefinition
* ApiDefinition 'name' was replaced with 'id'
* Api definition file was updated

### DIFF
--- a/storage/src/main/java/org/zalando/apidiscovery/storage/domain/model/ApiList.java
+++ b/storage/src/main/java/org/zalando/apidiscovery/storage/domain/model/ApiList.java
@@ -13,6 +13,6 @@ import java.util.List;
 @Builder
 public class ApiList {
 
-    private List<Api> apis;
+    private List<Api.ApiMetaData> apis;
 
 }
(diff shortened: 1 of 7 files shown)


storage/src/main/java/org/zalando/apidiscovery/storage/domain/model/ApiList.java
```diff
@@ -13,6 +13,6 @@
 @Builder
 public class ApiList {
```
maschleg: please rename this ``Apis``we aligned on using plurals instead of List suffix.

fmueller:
👍

PR #171, opened 2017-06-23
Build swagger-ui from source and exclude dependencies

### LINKED ISSUE #170: Errors caused by dependency duplication in production
Selecting an API from the drop down yields this error:

> Error fetching API +252ms Error: Minified React error #119; visit http://facebook.github.io/react/docs/error-decoder.html?invariant=119 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

This is caused by https://cdnjs.cloudflare.com/ajax/libs/swagger-ui/3.0.16/swagger-ui-bundle.js including React as a transitive dependency which is now included on the page twice.

mfellner:
Closes #170.

* build swagger-ui from original source with webpack
* externalize several large dependencies to reduce build time and bundle size


swagger-ui-3/webpack.config.js
```diff
@@ -2,6 +2,7 @@ const path = require('path');
 const webpack = require('webpack');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const args = require('./webpack.args');
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
```
mfellner: This should be removed...

swagger-ui-3/webpack.config.js
```diff
@@ -47,6 +47,7 @@ module.exports = {
     }
   },
   plugins: [
+    new BundleAnalyzerPlugin(),
```
mfellner: The bundle analyser should be removed...

maschleg:
👍

PR #173, opened 2017-06-26
Fix CDP build and use pip3

mfellner:
* use pip3 instead of pip
* remove scm-source building ([done by CDP](https://pages.github.bus.zalan.do/continuous-delivery/cdp-docs/user-guide/index.html?highlight=scm%20source#scm-source))
* load config file for api-portal (temporary)


maxim-tschumak:
:+1:

PR #178, opened 2017-06-26
Don't copy scm-source in swagger-ui-3 Dockerfile

mfellner:
Don't copy scm-source in swagger-ui-3 Dockerfile.


maxim-tschumak:
:+1:

PR #181, opened 2017-06-27
Swagger UI upgrades

### LINKED ISSUE #180: Upgrade swagger-ui-3
* Upgrade npm dependencies
* Update Docker build to not rely on configuration file
* Add some documentation

mfellner:
* upgrade dependencies
* update tests
* support nested configuration options with environment variables
* change the CDP monorepo script to use the pr number environment variable if possible

Closes #180


swagger-ui-3/README.md
```diff
@@ -33,4 +26,40 @@ npm start
 
 ### Configuration
 
-All configuration is external to the application and can be set with configuration files, environment variables and command line arguments. Check the [/conf](/conf) directory for examples and take a look at [conf.ts](/server/framework/conf.ts) to see the implementation.
+All configuration is managed by [TypeConf](https://github.com/mfellner/typeconf) and can be provided through configuration files, environment variables and command line arguments. Here is a list of the supported options:
```
maschleg: nice lib :sunglasses:

swagger-ui-3/README.md
```diff
@@ -16,14 +10,13 @@ Then you can load new API definitions by typing in a valid Api Discovery Storage
 npm run dev
 ```
 
-Configure authentication by setting a valid OAuth token:
+Set the base URL of the api-storage service and provide a valid OAuth token by exporting these environment variables:
 
 ```sh
-export API_DISCOVERY_OAUTH2_ACCESS_TOKENS=api-storage-domain=some-token
+export API_DISCOVERY_OAUTH2_ACCESS_TOKENS="apistorage.com=some-token"
+export API_DISCOVERY_API_STORAGE_CONF__BASE_URL="https://apistorage.com"
```
maschleg: 1. please use API repository instead of storage as we will soon rename it.
2. I would use API_PORTAL instead of API_DISCOVERY as prefix, because otherwise I would assume this var belongs to the API discovery (crawler) application.

mfellner: OK, makes sense. My first thought was to use the name of the whole project/repository but it's better to use the service's name.

swagger-ui-3/common/domain/model/ClientOAuth2Conf.ts
```diff
@@ -41,8 +42,16 @@ export default class ClientOAuth2Conf implements ClientAuthConf {
...
       requestParameters?: object;
     },
-    urlResolver: (url: string) => string = _ => _
+    urlResolver: (url: string) => string = _ => _,
+    objectParser = (s?: any) => (typeof s === 'string' ? JSON.parse(s || '{}') : s)
   ) {
-    Object.assign(this, { ...data, clientId: urlResolver(data.clientId) });
+    super(
+      {
+        ...data,
```
maxim-tschumak: nice

.delivery/build-swagger-ui-3.sh
```diff
@@ -13,7 +13,7 @@ echo "Build api-portal."
 
 if [ "$IS_PR_BUILD" = true ]; then
```
maxim-tschumak: Should we take this check out of the build scripts? (either in python script, or a util bash script)

swagger-ui-3/Dockerfile
```diff
@@ -10,17 +10,17 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
-COPY yarn.lock /opt/api-discovery-ui/
+COPY package.json /opt/api-portal-ui/
+COPY yarn.lock /opt/api-portal-ui/
 
-WORKDIR /opt/api-discovery-ui
+WORKDIR /opt/api-portal-ui
 RUN yarn install --production && yarn cache clean
 
-COPY dist /opt/api-discovery-ui/
+COPY dist /opt/api-portal-ui/
```
maxim-tschumak: Please combine multiple `COPY` commands with the same destination to one. By this, we can decrease the number of layers.

.delivery/build-swagger-ui-3.sh
mfellner: Imo, the build.py script should just do the delegation. At some point we might want to run some tasks for a pull request (just not push the Docker image).

We could factor this out into a separate file but it's not a lot of code... I'd rather keep the bash scripts simple.

maxim-tschumak: OK, we can do this later.

maxim-tschumak:
:+1:

PR #182, opened 2017-06-28
Refactor api integration to component test

### LINKED ISSUE #116: Rewrite API integration tests into component tests
Since we changed our testing strategy, all api integration tests should be rewritten to component tests:

- server is not ramped up
- `MockMvc` is used
- `AbstractComponentTest` can be used as base class (available once #110 is done)

maschleg:
closes #116


storage/src/test/java/org/zalando/apidiscovery/storage/domain/service/ApiDefinitionProcessingServiceComponentTest.java
```diff
@@ -1,22 +1,47 @@
...
+    private ApiRepository apiRepository;
+
+    @Autowired
+    private EntityManager entityManager;
+
     private ApiDefinitionProcessingService apiService;
 
+    @Before
+    public void setUp() throws Exception {
+        apiService = new ApiDefinitionProcessingService(applicationRepository, apiRepository, entityManager);
```
fmueller: This is not necessary: Spring automatically injects the necessary fields  if you put `@Autowired` to `apiService` field.

storage/src/test/java/org/zalando/apidiscovery/storage/repository/ApiRepositoryTest.java
```diff
@@ -11,7 +17,17 @@
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class ApiRepositoryTest extends AbstractComponentTest {
+@RunWith(SpringRunner.class)
+@DataJpaTest
+@AutoConfigureTestDatabase(replace = Replace.NONE)
+@Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:cleanDatabase.sql")
```
fmueller: Maybe it would be better to refactor all these database test related annotations into a `AbstractDatabaseTest`. Otherwise if we change sth later, we would have to change all database related tests.

storage/src/test/java/org/zalando/apidiscovery/storage/AbstractComponentTest.java
```diff
@@ -56,12 +56,6 @@ public MockMvc mockMvc(final WebApplicationContext context) throws Exception {
         }
     }
 
-    protected ResultActions postApiDefinition(Resource apiDefinition) throws Exception {
-        return mvc.perform(post("/api-definitions")
-            .contentType(APPLICATION_JSON_UTF8_VALUE)
-            .content(TestDataHelper.readResource(apiDefinition)));
-    }
-
```
mfellner: There are several unused imports in this class now that we should remove.

storage/src/test/java/org/zalando/apidiscovery/storage/domain/service/ApiDefinitionProcessingServiceComponentTest.java
```diff
@@ -1,22 +1,47 @@
...
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.zalando.apidiscovery.storage.TestDataHelper.discoveredMetaApi;
 
-public class ApiDefinitionProcessingServiceIntegrationTest extends AbstractResourceIntegrationTest {
+@RunWith(SpringRunner.class)
+@DataJpaTest
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:cleanDatabase.sql")
+public class ApiDefinitionProcessingServiceComponentTest {
```
mfellner: Can we extend `AbstractComponentTest` here?

storage/src/test/java/org/zalando/apidiscovery/storage/domain/service/ApiDefinitionProcessingServiceComponentTest.java
mfellner: Yes, but is the Spring context recreated between tests? Or is the ApiDefinitionProcessingService instance going to stay the same? It shouldn't matter in this case but it would still be interesting to know...

storage/src/test/java/org/zalando/apidiscovery/storage/repository/ApiRepositoryTest.java
mfellner: What about `AbstractComponentTest`?

storage/src/test/java/org/zalando/apidiscovery/storage/resource/ApiResourceComponentTest.java
```diff
@@ -143,16 +136,13 @@ public void shouldReturnAllInActiveApis() throws Exception {
...
+            .andExpect(status().isOk())
+            .andExpect(jsonPath("$.apis", hasSize(1)))
+            .andExpect(jsonPath("$.apis[0].id", equalTo(ANOTHER_API)))
+            .andExpect(jsonPath("$.apis[0].lifecycle_state", equalTo(INACTIVE)));
     }
 
-
     @Test
     public void shouldReturnAllDecommissionedApis() throws Exception {
         ApplicationEntity app1 = givenApplication(APP1);
```
mfellner: We have a lot of this setup code in different styles and places, e.g. org/zalando/apidiscovery/storage/repository/ApiRepositoryTest.java:34.

It would be a good idea to factor this code out into some kind of functional fixture utility class. Maybe something like this:

```java
static Function<String, ApplicationEntity> applicationRepositoryInitializer(ApplicationRepository applicationRepository) {
    return (String name) -> applicationRepository.save(
        ApplicationEntity
            .builder()
            .name(name)
            .appUrl("/info")
            .created(now(UTC))
            .build()
    );
}
```
Or some other pattern, I don't know (Java is ugly 😞 ).

mfellner: It would also be cleaner, imo, if we put *all* of our fixture constants into a Fixture.class as well. @tkrop?

storage/src/test/java/org/zalando/apidiscovery/storage/repository/ApiRepositoryTest.java
maschleg: The current AbstractComponentTest configures a web context, which isn't necessary for this test case. I would suggest the following AbstractComponentTests:

``AbstractResourceComponentTest``: includes Resource, Service and Repository (embedded DB) layers 
``AbstractServiceComponentTest``: includes Service and Repository (embedded DB) layers 
``AbstractDatabaseComponentTest``: includes Repository (embedded DB) layer

storage/src/test/java/org/zalando/apidiscovery/storage/domain/service/ApiDefinitionProcessingServiceComponentTest.java
maschleg: This test is without component scan, so I have to instantiate the service in this class.

storage/src/test/java/org/zalando/apidiscovery/storage/resource/ApiResourceComponentTest.java
maschleg: Yes that's true we should have a uniform style to create domain objects. However, we already had a discussion and decided not to have a common Fixture.class or DomainObjectGen.class, because changes in this class will then have impact on all test classes. So every test class should be responsible for it's own fixture/domain creation.

storage/src/test/java/org/zalando/apidiscovery/storage/AbstractComponentTest.java
maschleg: should be removed now.

maschleg:
@fmueller @mfellner please review changes

mfellner:
👍

fmueller:
👍

PR #185, opened 2017-06-30
Final API Portal UI Proposal (1st iteration)

### LINKED ISSUE #155: API Discovery User Interface
### What
Design and implement a user interface for Swagger UI that makes it easy to discover a particular API definition.

### Why
The current implementation of a drop-down list of available APIs has several drawbacks:

* not enough information to find an API (names are derived from IDs)
* is not searchable
* generally bad UX when the list is long
* does not scale to higher numbers of APIs

### Requirements
A new user interface for API discovery should serve the following needs:

* users can find an API by the title of the Open API definition
* users are not overwhelmed by the number of available API definitions
* the interface can support future UI refinements, e.g. filtering by certain attributes, displaying multiple versions of an API definition, etc.

### Proposal
**API fuzzy search interface.** A large search box allows users to enter text (e.g. the title of the API definition) in order to filter a list of at most `N` results sorted by criterion `C` that match the entered text.

* ✅  can be naively implemented as full-text search on API definitions
* ✅  as the user enters more text, the number of search results quickly drops
* ✅  users who know what they are looking for are getting results fast
* ⚠️  full-text search may be slow
* ⚠️  full-text search may yield ambiguous results
* ⚠️  no overview of total collection of API definitions

mfellner:
Closes #155 

This is the proposal for the final API portal UI (in it's first iteration):

* api version selector
* definition selector
* app view

screen shot 2017-06-30 at 12 19 36


fmueller:
I like the concept. :-)

The metadata about when an API was crawled per application and lifecycle state is missing. I think, we should add this information as well.

mfellner:
@fmueller I'm not sure we have this information in our current data model. What we do have is some temporal information about an application (I believe when it was **first discovered**) and temporal information about the **relations** between entities. Take a look at this overview:

model-overview

maschleg:
Nice Proposal.

- It is hard to get what definition means without knowing the data model. I would rename it to revision (changes on a smaller scale).  
- it is inconsistent that the version drop down ist not labeled but the definition drop down.

What about a multi level drop down? First level is the version and the second the revision. (e.g. https://bootsnipp.com/snippets/9naj). The selected "version" I would display version-revision e.g. 0.1.0-2

fmueller:
We know per api deployment when it was last time crawled. An api deployment is basically temporal manifestation of an API (one API version with one API definition) on an application.

So, the information is there. The question is if we wanna display it at all and when yes, how we wanna display it.

